PR #7, opened 2022-01-25
[PP][SUPP-64] cleanup

pawelpanasewicz:

### DIFF
--- a/app/moveittocor/corcommon/model/Journey.scala
+++ b/app/moveittocor/corcommon/model/Journey.scala
@@ -14,24 +14,24 @@
  * limitations under the License.
  */
 
-package models
+package moveittocor.corcommon.model
 
-import enumeratum.{ Enum, _ }
-import moveittocor.corcommon.internal.jsonext.EnumFormat
-import play.api.libs.json.Format
+import play.api.libs.json.{Json, OFormat}
 
-import scala.collection.immutable
+import java.time.Instant
 
-sealed trait JourneyStatus extends EnumEntry
+final case class Journey(
+  _id: JourneyId,
+  createdDate: Instant,
+  lastUpdated: Instant = Instant.now,
+  status: JourneyStatus,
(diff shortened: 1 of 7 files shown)


app/moveittocor/corcommon/model/Journey.scala
```diff
@@ -14,19 +14,22 @@
...
   _id: JourneyId,
   createdDate: Instant,
   lastUpdated: Instant = Instant.now,
-  status: JourneyStatus) {
+  status: JourneyStatus,
+  amount: Option[AmountInPence]
+  ) {
   def id: JourneyId = _id
+
+  def getAmount: AmountInPence = amount.getOrElse(throw new RuntimeException("amount should be there"))
```
cavster, line 32: If Amount should be there why have it as an option?

nubz: actually I don't think `Amount` or `def getAmount` should be in this model at this point in time, my guess is these examples were left in by accident as it won't compile like this either - @pawelpanasewicz is that right, I remember you demonstrating the type of thing we could use Journey for with these example additions?

pawelpanasewicz: The `amount` is optional in the journey because it's provided on creens by users. So a journey starts, with amount being None, later when user provides it, it becomes Some(123). Later in the application the amount would need to be used and taken out from Option[Amount] - for that purpose we have convenience method `getAmount` which expects that the amount is defined.

pawelpanasewicz: And yes - at the moment we don't really need this chunk of model. Its solely purpose is a demonstration of usage. Let's leave it please, it will be refactored anyway once we get clear requirements. For now we're just preparing repo for future development.

cavster: Cool

PR #14, opened 2022-03-31
SUPP-218[DM] Verify only users with EPAYE enrolment can access your_bill_is page

jamesMaxwell191:


app/actions/AuthenticatedAction.scala
```diff
@@ -51,19 +52,23 @@ class AuthenticatedAction @Inject() (
...
+        case enrolments ~ credentials if enrolments.getEnrolment("IR-PAYE").isDefined =>
           Future.successful(
             Right(new AuthenticatedRequest[A](request, enrolments, credentials)))
+        case _ =>
+          Future.failed(new IllegalStateException("not enrolled"))
       }.recover {
-        case _: NoActiveSession =>
-          //TODO: what is a proper value to origin
-          Left(Redirect(appConfig.loginUrl, Map("continue" -> Seq(appConfig.frontendBaseUrl + request.uri), "origin" -> Seq("pay-online"))))
+        case _: NoActiveSession | _: IllegalStateException => loginPage
```
pawelpanasewicz, line 61: I'm not sure if we wanted to send user to login page if he doesn't have enrolments. 
Probably we should send him to the page instructing him that he needs to setup EPAYE enrolment. 

I'd catch in above pattern matching case when user is logged in and has no required enrolments and redirect him to a a new page "youre not enorlled". We will chase BA and content designers to design that page.

Just to remind, on the horizon are VAT and other tax regimes. Let's make it generic to be capable of dealing with new tax regimes.

jamesMaxwell191: Ok . I'll add support for the not enrolled page

jamesMaxwell191: I have added an abstraction for tax originator and tax regime.  I'm not sure if it is the right one. Time will tell.

nubz (approved):
LGTM

PR #23, opened 2022-04-21
[PP][SUPP-274] remove redundand config and move frontend specific urls to the correct place

pawelpanasewicz:

### DIFF
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -14,7 +14,7 @@
 
 include "frontend.conf"
 
-appName="essttp-frontend"
+appName = "essttp-frontend"
 
 play.http.router = prod.Routes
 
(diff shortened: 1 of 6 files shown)


conf/application.conf
```diff
@@ -38,60 +38,47 @@ play.modules.enabled += "uk.gov.hmrc.mongo.play.PlayMongoModule"
...
 
-    services {
-      auth {
-        protocol = http
-        host     = localhost
-        port     = 8500
-      }
-
-      auth-login-stub {
-            protocol = http
```
pawelpanasewicz, line 58: we don't define those urls here. Those are not (backend) "services" but just urls to other frontend apps.

RajuMandapati85 (approved):
Currently merging it as it's mainly changing the Configuration.

PR #26, opened 2022-04-27
[JB][SUPP-231/251]

barnesjake:
- add in generic ineligble page
   - add debt too large content for dynamic page
   - various warns removed

### DIFF
--- a/app/controllers/EPayeStartController.scala
+++ b/app/controllers/EPayeStartController.scala
@@ -18,14 +18,16 @@ package controllers
 
 import _root_.actions.Actions
 import essttp.rootmodel.TaxRegime
+import messages.{ Message, Messages }
 import models.{ EligibilityData, ttp }
-import play.api.i18n.{ I18nSupport, Messages }
+import play.api.i18n.I18nSupport
 import play.api.mvc._
 import services.{ EligibilityDataService, JourneyService }
+import testOnly.models.EligibilityError._
 import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendController
 import util.Logging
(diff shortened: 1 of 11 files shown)


pawelpanasewicz (approved):

app/controllers/EPayeStartController.scala
```diff
@@ -46,29 +48,38 @@ class EPayeStartController @Inject() (
...
-  val ePayeStartSubmit: Action[AnyContent] = as.default { implicit request =>
+  val ePayeStartSubmit: Action[AnyContent] = as.default { _ =>
     Redirect(routes.UpfrontPaymentController.upfrontPayment())
   }
 
   def routeResponse(data: EligibilityData)(implicit R: Request[_]): Result = {
     if (data.hasRejections) {
-      val errorMessage = data.rejections.map(_.entryName).mkString(", ")
-      Ok(errorTemplate("Error Page", "TTP Rejections", errorMessage)(implicitly[Request[_]], R.messages))
+      val leadingContentToShow: Seq[Message] = data.rejections.head match {
```
nubz, line 74: this method of using Messages alone might not be best for when we have more complex HTML structures with links, just commenting really as it obviously works for this use case, but we maybe could move to partials being included e.g DebtIsTooLarge.scala.html in later tickets if it proves a better way of handling the variable content

nubz (approved):

app/controllers/EPayeStartController.scala
barnesjake: Yeah I was thinking this, will refactor when the time comes

PR #27, opened 2022-04-27
[JB][SUPP-246] refactor to return html partials in the match statemen…

barnesjake:
…t, fix something I missed in the last commit


nubz (approved):
LGTM

PR #46, opened 2022-05-19
[AC][SUPP-302] dynamic service name + cleanup

AndyHWChung:


app/messages/Messages.scala
```diff
@@ -109,6 +109,10 @@ object Messages {
       english = "Set up a payment plan"
     )
 
+    val `Set up an Employers' PAYE payment plan`: Message = Message(
+      english = "Set up an Employers’ PAYE payment plan"
```
barnesjake, line 113: Does this need to be one of those "smart quote" things? I'm not 100% sure so disregard if not, but I've been caught out by this in the past

barnesjake (dismissed):

app/messages/Messages.scala
AndyHWChung: I made sure the message itself used a smart quote, but for the val name I chose a straight quote cos it's easier to type. Not that it's likely we're going to type the val name, but you never know...

nubz (dismissed):

app/connectors/TtpConnector.scala
```diff
@@ -22,7 +22,7 @@ import play.api.mvc.RequestHeader
 import requests.RequestSupport._
 import uk.gov.hmrc.http.HttpClient
 import uk.gov.hmrc.play.bootstrap.config.ServicesConfig
-import uk.gov.hmrc.http.HttpReads.Implicits
+import uk.gov.hmrc.http.HttpReads.Implicits._
```
pawelpanasewicz, line 25: Please mind the `HttpReadsUnitThrowingException` when importing that wildcard.

https://github.com/hmrc/essttp-backend/blob/main/cor-journey/src/main/scala/essttp/utils/HttpReadsUnitThrowingException.scala

app/views/templates/Layout.scala.html
```diff
@@ -90,31 +93,41 @@
...
-  @govTitle = @{serviceNameAndHeading + " - GOV.UK"}
+    @serviceName = @{
+        (request: AnyRef) match {
+          case r: JourneyRequest[_] =>
+              r.journey.taxRegime match {
+                  case TaxRegime.Epaye => Messages.ServicePhase.`Set up an Employers' PAYE payment plan`.show
+                  case TaxRegime.Vat => sys.error("VAT not handled yet")
+              }
+
+          case _ =>
```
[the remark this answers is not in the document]
pawelpanasewicz: Maybe safer would be to create dedicated `Layout` template for test-only pages which would accept `implicit request: Request[_]` and this view could be refactored to accept `implicit request: JourneyRequest[_]`?

pawelpanasewicz (dismissed):

app/connectors/TtpConnector.scala
AndyHWChung: :+1:  I've made the import more precise [here](https://github.com/hmrc/essttp-frontend/pull/46/commits/f21a1447f5c73e9ae1ba9ab708c055f6ccb285d7) now

barnesjake (approved):

app/views/templates/Layout.scala.html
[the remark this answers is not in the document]
AndyHWChung: I thought about doing that but it seemed like a lot of duplication - but I think I may be able to do it in a way which actually isn't a lot of duplication. I'll have a go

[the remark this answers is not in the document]
AndyHWChung: I've implemented a separate Layout [here](https://github.com/hmrc/essttp-frontend/pull/46/commits/c103abb42016264b804852302b4724a7e1c178bf)

app/connectors/TtpConnector.scala
AndyHWChung: to future proof I've just made sure the default `readUnit` is not brought in [here](https://github.com/hmrc/essttp-frontend/pull/46/commits/3eb80cd751658a9af7193e75f0e28e6b49ef0ef0)

AndyHWChung:
@barnesjake @nubz @pawelpanasewicz I've made a couple of changes after the comments above - do you mind looking at the PR again please? I've made a separate layout for the test-only stuff so that it's safer. 

I've ignore a test for now but that it something that is currently being worked on in a separate ticket so it will be replaced with a proper implementation soon anyway.

I've also made the logs nicer to look at:

Screenshot from 2022-05-19 15-40-56

Same info, but new colours to help read it better.

app/util/JourneyLogger.scala
```diff
@@ -90,11 +89,6 @@ object JourneyLogger {
     }
   }
 
-  private def makeRichMessage(message: => String, journeyId: JourneyId)(implicit request: RequestHeader): String = {
```
barnesjake, line 93: Do we need to remove this?

barnesjake (approved):

nubz (approved):

PR #47, opened 2022-05-20
Jake/supp 308

barnesjake:


app/controllers/UpfrontPaymentController.scala
```diff
@@ -17,53 +17,72 @@
...
+      case j: Journey.Stages.AfterComputedTaxId => logErrorAndRouteToDefaultPage(j)
+      case j: Journey.HasEligibilityCheckResult => displayPage(j)
+    }
   }
 
-  val upfrontPaymentSubmit: Action[AnyContent] = as.authenticatedJourneyAction.async { implicit request =>
-    upfrontPaymentForm()
+  private def displayPage(journey: Journey.HasEligibilityCheckResult)(implicit request: Request[_]): Result = {
+    val backUrl = journey.origin match {
+      case Origins.Epaye.Bta         => Some(routes.YourBillController.yourBill().url)
```
AndyHWChung, line 57: do we need a pattern match here if the backUrl is always the same thing?

app/controllers/UpfrontPaymentController.scala
```diff
@@ -87,23 +106,24 @@ class UpfrontPaymentController @Inject() (
...
-    Future.successful(Ok(upfrontSummaryPage(mockUserAnswers, AmountInPence(200000L))))
+    Future.successful(Ok(views.upfrontSummaryPage(mockUserAnswers, AmountInPence(200000L))))
   }
 }
 
 object UpfrontPaymentController {
-  val key: String = "UpfrontPaymentAmount"
 
-  def upfrontPaymentForm(): Form[String] = Form(
+  def canYouMakeAnUpfrontPaymentQuestionForm: Form[String] = Form(
```
AndyHWChung, line 115: we wouldn't need `formToCanPayUpfront` if we did some sort of transformation here and had `Form[Boolean]` or `Form[CanPayUpfront]`instead.

app/controllers/UpfrontPaymentController.scala
```diff
@@ -17,53 +17,72 @@
...
+        (validFormYesOrNo: String) => {
+          val canPayUpfront: CanPayUpfront = UpfrontPaymentController.formToCanPayUpfront(validFormYesOrNo)
+          val pageToRedirectTo: Call =
+            if (canPayUpfront.value) {
+              routes.UpfrontPaymentController.upfrontPaymentAmount()
+            } else {
+              routes.MonthlyPaymentAmountController.monthlyPaymentAmount()
+            }
+          journeyService.updateCanPayUpfront(request.journeyId, canPayUpfront)
+            .flatMap(_ => Future.successful(Redirect(pageToRedirectTo)))
```
AndyHWChung, line 78: you don't need a `flatMap` if you take out the `Future.successful` inside

app/services/JourneyService.scala
```diff
@@ -0,0 +1,52 @@
...
+  def updateEligibilityCheckResult(journeyId: JourneyId, eligibilityCheckResult: EligibilityCheckResult)(
+      implicit
+      requestHeader: RequestHeader
+  ): Future[Unit] = {
+    journeyConnector.updateEligibilityCheckResult(journeyId, eligibilityCheckResult)
+  }
+
+  object UpdateTaxRef {
+    //paye
+    def updatePayeTaxId(journeyId: JourneyId, empRef: EmpRef)(implicit requestHeader: RequestHeader): Future[Unit] = {
```
AndyHWChung, line 41: wouldn't `Epaye` instead of `Paye` be more accurate?

test/controllers/UpfrontPaymentControllerSpec.scala
```diff
@@ -21,22 +21,66 @@ import play.api.mvc.Result
...
 class UpfrontPaymentControllerSpec extends ItSpec {
 
   private val controller: UpfrontPaymentController = app.injector.instanceOf[UpfrontPaymentController]
 
-  "GET /upfrontpayment or whatever - Pawel, shall we add tests like these or just stick with int?" - {
-    "return 200 and the upfront payment page" ignore {
-      val fakeRequest = FakeRequest("GET", "/upfront-payment")
-      val result: Future[Result] = controller.upfrontPayment(fakeRequest)
+  "GET /can-you-make-an-upfront-payment" - {
+    "return 200 and the can you make an upfront payment page" in {
```
AndyHWChung, line 36: what do you think about also testing to see if nothing is selected on the form if there isn't an existing answer in session and also testing to see that the correct thing is selected if there is an existing answer in session?

Other possible things to test are also the back link location or the href of the submit button.

test/testsupport/stubs/EssttpBackend.scala
```diff
@@ -0,0 +1,235 @@
...
+        }
+      stubFor(
+        get(urlPathEqualTo(findByLatestSessionIdUrl))
+          .willReturn(aResponse()
+            .withStatus(200)
+            .withBody(jsonBody(additionalPayloadInfo)))
+      )
+    }
+
+    val jsonBody: ((String, Boolean)) => String = (tuple: (String, Boolean)) =>
```
AndyHWChung, line 149: you could make this easier to read by making it a `def like:
```scala
def jsonBody(blah: String, blahblah: Boolean): String = ...
```
with suitable names for the input parameters

app/controllers/UpfrontPaymentController.scala
barnesjake: true, will simplify

app/services/JourneyService.scala
barnesjake: will update

app/controllers/UpfrontPaymentController.scala
barnesjake: will make this change now

pawelpanasewicz: This one is tricky and important.

We don't want to rely on play's default messages which requires us to provide some translations in messages files. We'd be more happy if we could pass required messages directly into form formatters, like here:
https://github.com/hmrc/essttp-frontend/blob/636ba084caebf665a4ffad0a9383666d8ae58d08/app/testOnly/formsmodel/StartJourneyForm.scala#L42

So we need to figure out how to pass required messages to the form validator if the field is missing. 
I'll create an example in a moment.

pawelpanasewicz: https://github.com/hmrc/essttp-frontend/pull/48

app/controllers/UpfrontPaymentController.scala
```diff
@@ -17,53 +17,78 @@
...
+      case j: Journey.Stages.AfterStarted       => logErrorAndRouteToDefaultPage(j)
+      case j: Journey.Stages.AfterComputedTaxId => logErrorAndRouteToDefaultPage(j)
+      case j: Journey.HasEligibilityCheckResult => displayPage(j)
+    }
   }
 
-  val upfrontPaymentSubmit: Action[AnyContent] = as.authenticatedJourneyAction.async { implicit request =>
-    upfrontPaymentForm()
+  private def displayPage(journey: Journey.HasEligibilityCheckResult)(implicit request: Request[_]): Result = {
+    val backUrl = journey.origin match {
```
AndyHWChung, line 62: wouldn't it be simpler to have something like this?
```scala
val back = routes.YourBillController.yourBill()
Ok(views.canYouMakeAnUpFrontPayment(CanPayUpfrontForm.form, Some(back.url)))
```

test/controllers/UpfrontPaymentControllerSpec.scala
```diff
@@ -21,22 +21,105 @@ import play.api.mvc.Result
...
+      val result: Future[Result] = controller.canYouMakeAnUpfrontPayment(fakeRequest)
 
       status(result) shouldBe Status.OK
       contentType(result) shouldBe Some("text/html")
       charset(result) shouldBe Some("utf-8")
-      contentAsString(result) should include("Can you make an upfront payment?")
+
+      val pageContent: String = contentAsString(result)
+
+      pageContent should include(serviceName)
```
AndyHWChung, line 59: we've got Jsoup already added as a dependency. We could have stronger tests here like:
```scala
val pageContent: String = contentAsString(result)
val doc = Jsoup.parse(pageContent)

doc.select("#back").attr("href") shouldBe routes.WhereEver.Itis.url
doc.select("form").attr("action") shouldBe routes.UpfrontPaymentControllerSpec.canPayUpfrontSubmit.url
doc.select(""selector-for-p1").text shouldBe "p1 text"
etc.
```

app/controllers/UpfrontPaymentController.scala
barnesjake: oops sorry forgot that one will do now

test/controllers/UpfrontPaymentControllerSpec.scala
barnesjake: Oh that's nicer, will implement

test/controllers/UpfrontPaymentControllerSpec.scala
```diff
@@ -16,27 +16,112 @@
...
+
+    val pageContent: String = contentAsString(result)
+    val doc: Document = Jsoup.parse(pageContent)
+
+    doc.title() shouldBe s"Error $expectedPageTitle"
+    doc.select(".govuk-fieldset__heading").text() shouldBe expectedH1
+    doc.select(".hmrc-header__service-name").text() shouldBe expectedServiceName
+    doc.select(".hmrc-sign-out-nav__link").attr("href") shouldBe "http://localhost:9949/auth-login-stub/session/logout"
+    doc.select("#CanYouMakeAnUpFrontPayment-hint").text() shouldBe expectedPageHint
+    doc.select("#main-content > div > div > div > div > div > ul > li > a").text() shouldBe "Select yes if you can make an upfront payment"
```
AndyHWChung, line 124: a nicer way to select this would be something like this:
```scala
 val errorSummary = doc.select(".govuk-error-summary")
 val errorLink = errorSummary.select("a")
 errorLink.text() shouldBe "Select blah..."
 errorLink.attr("href") shouldBe "#CanYouMakeBlah"
```

AndyHWChung (approved):

PR #48, opened 2022-05-20
[PP][SUPP-308] how to messages in forms

pawelpanasewicz:


app/controllers/UpfrontPaymentController.scala
```diff
@@ -112,18 +121,37 @@ class UpfrontPaymentController @Inject() (
...
+  object CanMakeUpfrontPayment extends Enum[CanMakeUpfrontPayment] {
+    case object Yes extends CanMakeUpfrontPayment
+    case object No extends CanMakeUpfrontPayment
+    override def values: immutable.IndexedSeq[CanMakeUpfrontPayment] = findValues
+  }
+
+  def form(implicit language: Language): Form[CanMakeUpfrontPayment] = {
+
+    val canMakeUpfrontPaymentMapping: Mapping[CanMakeUpfrontPayment] = Forms.of(EnumFormatter.format(
+      enum                    = CanMakeUpfrontPayment,
```
pawelpanasewicz, line 139: This mapping accepts messages.

app/messages/Messages.scala
```diff
@@ -245,9 +245,9 @@ object Messages {
       english = "No"
     )
 
-    def getError(key: String): Message = key match {
-      case "error.required" => Message(english = "Select yes if you can make an upfront payment")
-    }
+    val `Select yes if you can make an upfront payment`: Message = Message(
```
pawelpanasewicz, line 248: We don't need to rely on a `String` key to get access to typesafe message

app/views/CanYouMakeAnUpfrontPayment.scala.html
```diff
@@ -52,7 +53,7 @@
                 ErrorSummary(errorList = form.errors.map(e =>
                     ErrorLink(
                         href = Some(s"#${e.key}"),
-                        content = Text(Messages.UpfrontPayment.getError(e.message).show)
+                        content = Text(e.message)
```
pawelpanasewicz, line 56: Here `e.message` contains localised message which was defined in form mapping.

app/views/CanYouMakeAnUpfrontPayment.scala.html
```diff
@@ -85,7 +86,7 @@
                     )
                 ),
                 classes = "govuk-radios--inline",
-                errorMessage = form.error(key).map(e => ErrorMessage(content = Text(Messages.UpfrontPayment.getError(e.message).show)))
+                errorMessage = form.error(key).map(e => ErrorMessage(content = Text(e.message)))
```
pawelpanasewicz, line 89: Again, here a `e.message` contains localised message which was defined in form mapping.

app/controllers/UpfrontPaymentController.scala
```diff
@@ -112,18 +121,37 @@ class UpfrontPaymentController @Inject() (
 
 object UpfrontPaymentController {
 
-  def canYouMakeAnUpfrontPaymentQuestionForm: Form[String] = Form(
-    mapping(
-      "CanYouMakeAnUpFrontPayment" -> nonEmptyText
-    )(identity)(Some(_))
-  )
+  sealed trait CanMakeUpfrontPayment extends enumeratum.EnumEntry {
```
pawelpanasewicz, line 124: Some basic model which will be used in form.

app/views/CanYouMakeAnUpfrontPayment.scala.html
```diff
@@ -75,17 +76,17 @@
                 items = Seq(
                     RadioItem(
                         content = Text(Messages.UpfrontPayment.`Yes`.show),
-                        value = Some("Yes"),
-                        checked = form.data.values.toList.contains("Yes")
+                        value = Some(CanMakeUpfrontPayment.Yes.entryName),
```
pawelpanasewicz, line 79: Instead of strings we can rely on the actual objects

app/controllers/UpfrontPaymentController.scala
AndyHWChung: the name here can get confusing with the root `CanPayUpfront`model - maybe we could adopt some convention like `CanPayUpfrontFormValue` as a name for this sealed trait here and beyond?

pawelpanasewicz: Yes, that makes sense.

pawelpanasewicz: Done

app/controllers/UpfrontPaymentController.scala
```diff
@@ -112,18 +121,37 @@ class UpfrontPaymentController @Inject() (
 
 object UpfrontPaymentController {
 
-  def canYouMakeAnUpfrontPaymentQuestionForm: Form[String] = Form(
-    mapping(
-      "CanYouMakeAnUpFrontPayment" -> nonEmptyText
-    )(identity)(Some(_))
-  )
+  sealed trait CanUpfrontPaymentFormValue extends enumeratum.EnumEntry {
```
AndyHWChung, line 124:
```suggestion
  sealed trait CanPayUpfrontFormValue extends enumeratum.EnumEntry {
```
that goes a bit better with the `CanPayUpfront` root model

app/controllers/UpfrontPaymentController.scala
AndyHWChung: :+1: thanks

PR #50, opened 2022-05-26
Jake/supp 309

barnesjake:


pawelpanasewicz (approved):

app/controllers/UpfrontPaymentController.scala
```diff
@@ -77,46 +81,77 @@ class UpfrontPaymentController @Inject() (
...
-            .map(_ => Redirect(pageToRedirectTo))
+            .map(_ => Redirect(pageToRedirectTo.url))
         }
       )
   }
 
-  val upfrontPaymentAmount: Action[AnyContent] = as.authenticatedJourneyAction.async { implicit request =>
-    val mockJourney = MockJourney(userAnswers = UserAnswers.empty.copy(hasUpfrontPayment = Some(true)))
-    Future.successful(Ok(views.upfrontPaymentAmountPage(upfrontPaymentAmountForm(mockJourney), mockJourney.qualifyingDebt, AmountInPence(100L))))
+  val upfrontPaymentAmount: Action[AnyContent] = as.eligibleJourneyAction { implicit request =>
```
AndyHWChung, line 89: do we want an action for this endpoint which ensures that the user said they were able to make an upfront payment? Or was the idea to have the actions check as far as the eligibility only?

app/controllers/UpfrontPaymentController.scala
```diff
@@ -77,46 +81,77 @@ class UpfrontPaymentController @Inject() (
...
+    }
+  }
+
+  private val minimumUpfrontPaymentAmount: AmountInPence = appConfig.JourneyVariables.minimumUpfrontPaymentAmountInPence
+
+  private def displayUpfrontPageAmountPage(journey: Journey)(implicit request: Request[_]): Result = {
+    val backUrl: Option[String] = Some(routes.UpfrontPaymentController.canYouMakeAnUpfrontPayment().url)
+    val debtTotalAmount: DebtTotalAmount = UpfrontPaymentController.determineMaxDebt(journey) match {
+      case Some(value) => value
+      case None        => Errors.throwBadRequestException("I am an error, this should never happen")
```
AndyHWChung, line 103: `displayUpfrontPageAmountPage` is called at a point where a `Journey.AfterAnsweredCanPayUpfront` is available. If you make `UpfrontPaymentController.determineMaxDebt` take in that type instead does that get rid of one of the ways an error can occur?

In the error message is actually seen a more useful message would be better - something like `"Could not determine max debt"`?

app/controllers/UpfrontPaymentController.scala
```diff
@@ -77,46 +81,77 @@ class UpfrontPaymentController @Inject() (
...
-
-  def upfrontPaymentAmountForm(journey: MockJourney): Form[BigDecimal] = Form(
-    mapping(
-      "UpfrontPaymentAmount" -> Forms.of(amountOfMoneyFormatter(AmountInPence(100L).inPounds > _, journey.qualifyingDebt.inPounds < _))
-    )(identity)(Some(_))
-  )
+  def determineMaxDebt(journey: Journey): Option[DebtTotalAmount] = journey match {
+    case _: Journey.BeforeEligibilityChecked => None
+    case j: Journey.AfterEligibilityChecked =>
+      Some(j.eligibilityCheckResult.chargeTypeAssessment.map(_.debtTotalAmount)
```
AndyHWChung, line 154: this would make more sense as just
 ```scala
j.eligibilityCheckResult.chargeTypeAssessment.map(_.debtTotalAmount).headOption
```
The `Option` in the return type signifies the error case will be returned as a `None` rather than an exception being thrown. The place calling this is already handling the `None` case.

File: app/models/forms/UpfrontPaymentAmountForm.scala [hunk 4]
```diff
@@ -0,0 +1,45 @@
...
+import essttp.journey.model.Journey
+import essttp.journey.model.ttp.DebtTotalAmount
+import essttp.rootmodel.AmountInPence
+import essttp.utils.Errors
+import models.MoneyUtil.amountOfMoneyFormatter
+import play.api.data.Forms.mapping
+import play.api.data.{Form, Forms}
+
+object UpfrontPaymentAmountForm {
+  def form(journey: Journey, minimumPaymentAmount: AmountInPence): Form[BigDecimal] = {
```
AndyHWChung, line 28: it feels like a `DebtTotalAmount` should be passed into this form rather than an entire `Journey` - the calling site is already handling the case where this amount can't be found so it doesn't need to be replicated here.

app/controllers/UpfrontPaymentController.scala
barnesjake: I think eligibility only, could open this up to discussion though

app/controllers/UpfrontPaymentController.scala
barnesjake: just implemented this, resolving once I've pushed

app/controllers/UpfrontPaymentController.scala
barnesjake: just implemented this, resolving once I've pushed

app/models/forms/UpfrontPaymentAmountForm.scala
barnesjake: just implemented this, resolving once I've pushed

test/controllers/UpfrontPaymentControllerSpec.scala
```diff
@@ -55,75 +58,201 @@ class UpfrontPaymentControllerSpec extends ItSpec {
...
+
+    forAll(
+      Table(
+        ("Scenario flavour", "form input", "expected error message"),
+        ("x > maximum debt", "30001", "How much you can pay upfront must be £3,000.00 or less"),
+        ("x < 1", "0.99", "How much you can pay upfront must be £1.00 or more"),
+        ("x < 0", "-1", "How much you can pay upfront must be £1.00 or more"),
+        ("x = 0", "0", "How much you can pay upfront must be £1.00 or more"),
+        ("x = NaN", "one", "How much you can pay upfront must be an amount of money"),
+        ("x = null", "", "How much you can pay upfront must be an amount of money")
```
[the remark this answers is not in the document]
barnesjake: yup, just corrected, I've updated the messages file to reflect this too

AndyHWChung (approved):

PR #55, opened 2022-05-27
[JB][SUPP-311] add upfront payment summary page, plus a few nice id's…

barnesjake:
… for raju :)


app/messages/Messages.scala
```diff
@@ -346,20 +346,31 @@ object Messages {
       english = "Payment summary"
     )
 
+    val `Can you make an upfront payment?`: Message = Message(
```
AndyHWChung, line 349: the question and answer should always be the same as the title and answer on the actual page - on the summary page template then you should be able to just reuse what's in the `UpfrontPayment` object in this file

app/views/UpfrontPaymentSummary.scala.html
```diff
@@ -0,0 +1,112 @@
...
+    </h1>
+    @govukSummaryList(SummaryList(
+        rows = Seq(
+            SummaryListRow(
+                key = Key(
+                    content = HtmlContent(Html(Messages.UpfrontPaymentSummary.`Can you make an upfront payment?`.show)),
+                    classes = "govuk-!-width-one-half"
+                ),
+                value = Value(
+                    content = Text(Messages.UpfrontPaymentSummary.`Yes`.show)
```
AndyHWChung, line 62: the value here is dependent on what the user actually chose - is there a way to make the value here dependent on what's in session rather being hard-coded?

AndyHWChung: or do you reach this summary page if you said you can't make an upfront payment?

barnesjake: It's only reachable if they select yes for upfront payment so I thought it's easier to hard code as yes

app/messages/Messages.scala
barnesjake: Sorry I don't understand

barnesjake: Oh wait I think I do, sorry

AndyHWChung: the same holds for the `Yes` message as well - there's one above you can use already

barnesjake: updated

app/views/UpfrontPaymentSummary.scala.html
```diff
@@ -0,0 +1,112 @@
...
+)
+
+@import requestSupport._
+
+
+@title = @{
+    Messages.UpfrontPaymentSummary.`Payment summary`.show
+}
+
+@layout(pageTitle = Some(title), showSignOutLink = false, backUrl = backUrl) {
```
AndyHWChung, line 49: shouldn't we be showing the sign out link on every page in the journey once you're logged in?

barnesjake: Good spot, it wasn't actually used in the template so removing it

AndyHWChung (approved):

PR #72, opened 2022-07-05
[JB][SUPP-159] add enter direct debit details page

barnesjake:


app/controllers/IneligibleController.scala
```diff
@@ -62,4 +63,9 @@ class IneligibleController @Inject() (
   val alreadyHaveAPaymentPlanPage: Action[AnyContent] = as.authenticatedJourneyAction { implicit request =>
     Ok(views.partials.ineligibleTemplatePage(Messages.NotEligible.`You already have a payment plan with HMRC`, views.partials.existingPaymentPlanPartial()))
   }
+
+  val cannotSetupDirectDebitOnlinePage: Action[AnyContent] = as.eligibleJourneyAction { implicit request =>
```
AndyHWChung, line 67: not sure if this belongs in the `ineligibleController` - they're eligible according to the rules but they just can't set up a direct debit with the bank details given. Maybe this could go in the bank details controller?

barnesjake: good point, I'll move it

app/models/forms/BankDetailsForm.scala
```diff
@@ -0,0 +1,79 @@
...
+
+    val isSoleSignatoryFormMapping: Mapping[IsSoleSignatoryFormValue] = Forms.of(EnumFormatter.format(
+      enum                    = IsSoleSignatoryFormValue,
+      errorMessageIfMissing   = "error.required",
+      errorMessageIfEnumError = "error.required"
+    ))
+
+    Form(
+      mapping(
+        "name" -> nonEmptyText(maxLength = 100),
```
AndyHWChung, line 71: do we need to worry about special characters or accented characters? We probably should be able to restrict the character set here a bit for a name

barnesjake: Might be one to check

app/controllers/IneligibleController.scala
barnesjake: done

test/controllers/BankDetailsControllerSpec.scala
```diff
@@ -0,0 +1,249 @@
...
+      val errorSummary = doc.select(".govuk-error-summary__list")
+      val errorLink = errorSummary.select("a").asScala.toList
+      errorLink(0).text() shouldBe "Enter the name on the account"
+      errorLink(0).attr("href") shouldBe accountNameFieldId
+      errorLink(1).text() shouldBe "Enter sort code"
+      errorLink(1).attr("href") shouldBe sortCodeFieldId
+      errorLink(2).text() shouldBe "Enter account number"
+      errorLink(2).attr("href") shouldBe accountNumberFieldId
+      errorLink(3).text() shouldBe "Select yes if you are the account holder"
+      errorLink(3).attr("href") shouldBe accountHolderRadioId
```
AndyHWChung, line 185: there's a bit of repetition here you could probably condense down into a function like:
```scala
def testFormError(
  formInput: (String,String)*
)(
  expectedErrorTextWithHrefs: List[(String, String)]
)
```

barnesjake: I agree but felt it was a bit more readable that way, happy to change it though

AndyHWChung (approved):

PR #74, opened 2022-07-06
[JB][SUPP-161] add confirm dd details page

barnesjake:


app/controllers/BankDetailsController.scala
```diff
@@ -98,7 +98,19 @@ class BankDetailsController @Inject() (
   }
 
   val checkBankDetails: Action[AnyContent] = as.eligibleJourneyAction { implicit request =>
-    Ok("This is where the check bank details page will go, for now, here's the journey:\n\n\n" + Json.prettyPrint(Json.toJson(request.journey)))
+    request.journey match {
+      case j: Journey.BeforeEnteredDirectDebitDetails => JourneyIncorrectStateRouter.logErrorAndRouteToDefaultPage(j)
+      case j: Journey.AfterEnteredDirectDebitDetails  => displayCheckBankDetailsPage(j)
```
AndyHWChung, line 103: we could put an extra guard here on the stage to see if the user said that they were the sole signatory - if they said no then we shouldn't show this page

test/controllers/BankDetailsControllerSpec.scala
```diff
@@ -220,12 +237,64 @@ class BankDetailsControllerSpec extends ItSpec {
...
+      val summaries = doc.select(".govuk-summary-list").select(".govuk-summary-list__row").iterator().asScala.toList
+      summaries.size shouldBe 3
+
+      val expectedAccountNameRow = SummaryRow("Account name", "Bob Ross", routes.BankDetailsController.enterBankDetails().url)
+      val expectedSortCodeRow = SummaryRow("Sort code", "123456", routes.BankDetailsController.enterBankDetails().url)
+      val expectedAccountNumberRow = SummaryRow("Account number", "12345678", routes.BankDetailsController.enterBankDetails().url)
+      val expectedSummaryRows = List(expectedAccountNameRow, expectedSortCodeRow, expectedAccountNumberRow)
+      extractSummaryRows(summaries) shouldBe expectedSummaryRows
+
+      //assert direct debit content
```
AndyHWChung: is this something you forgot to do or intended to do in the future?

app/controllers/BankDetailsController.scala
barnesjake: Sure, will update

test/controllers/BankDetailsControllerSpec.scala
barnesjake: Yeah, I forgot to do this but already updated in most recent commit

nubz (approved):

app/controllers/BankDetailsController.scala
barnesjake: Done, I added a test for this too

app/controllers/BankDetailsController.scala
```diff
@@ -98,7 +98,20 @@ class BankDetailsController @Inject() (
...
+      case j: Journey.AfterEnteredDirectDebitDetails =>
+        if (j.directDebitDetails.isAccountHolder) {
+          Ok(views.bankDetailsSummary(j.directDebitDetails, BankDetailsController.enterBankDetailsUrl))
+        } else {
+          Redirect(routes.BankDetailsController.cannotSetupDirectDebitOnlinePage().url)
+        }
+    }
+  }
+
+  val checkBankDetailsSubmit: Action[AnyContent] = as.eligibleJourneyAction.async { implicit request =>
```
AndyHWChung, line 112: we should probably put that check on the stage on the submit endpoint as well - it'd be bad if we allowed someone to use this POST endpoint to bypass that check and proceed anyway. Not sure if anyone actually would but better to be safe

barnesjake: will raise a ticket, I assume you mean in the backend?

barnesjake: we'll probably be able to bring it into this sprint tbh

AndyHWChung: no I meant around the bit of code before we talk to the backend here

barnesjake: Done

AndyHWChung (approved):

PR #75, opened 2022-07-07
Jake/supp 164

barnesjake:


test/controllers/BankDetailsControllerSpec.scala
```diff
@@ -318,6 +323,67 @@ class BankDetailsControllerSpec extends ItSpec {
...
+        .map(things => things._1.text() shouldBe things._2)
+
+      val bulletListItems = List(
+        "pay late or miss a payment",
+        "pay another tax bill late",
+        "do not submit your future tax returns on time"
+      )
+      doc.select(".govuk-list--bullet").select("li").asScala.toList
+        .zip(bulletListItems)
+        .map(things => things._1.text() shouldBe things._2)
```
AndyHWChung, line 366: we could probably do a bit better than `things` as a name 😆 Maybe something like this?
```scala
 .map{ case (element, expectedText) => element.text() shouldBe expectedText }
```

barnesjake: haha as I was writing it I thought I should do better 😂 , will update

AndyHWChung (approved):

PR #76, opened 2022-07-08
[PP] private constructor so it's only possible to construct message using apply method

pawelpanasewicz:

### DIFF
--- a/app/messages/Message.scala
+++ b/app/messages/Message.scala
@@ -18,7 +18,7 @@ package messages
 
 import langswitch.{Language, Languages}
 
-final case class Message(
+final case class Message private (
     english: String,
     welsh:   Option[String]
 ) {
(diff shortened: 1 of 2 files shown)


app/messages/Message.scala
```diff
@@ -18,7 +18,7 @@ package messages
 
 import langswitch.{Language, Languages}
 
-final case class Message(
+final case class Message private (
```
AndyHWChung, line 21: this may not do what you think it does - I think you'd still be able to do this:
```scala
Message("english", Some("welsh"))
```
but you're just prevented from doing this:
```scala
new Message("english", Some("welsh"))
```
I'll have a double check...

AndyHWChung: yeah, I think that's the case after a little test locally

pawelpanasewicz: Yup, thanks for pointing this out. I think I've fixed that.

AndyHWChung (approved):

PR #78, opened 2022-07-08
Jake/supp 367

barnesjake:


app/controllers/LandingController.scala
```diff
@@ -17,7 +17,7 @@
 package controllers
 
 import _root_.actions.Actions
-import essttp.journey.model.Origins
+//import essttp.journey.model.Origins
```
AndyHWChung, line 20: this can be cleaned up if we don't need it anymore

app/controllers/LandingController.scala
```diff
@@ -33,13 +33,12 @@ class LandingController @Inject() (
...
-    request.journey.origin match {
-      case Origins.Epaye.Bta         => Ok(views.epayeLanding())
-      case Origins.Epaye.DetachedUrl => Ok(views.epayeLanding())
-      case Origins.Epaye.GovUk       => Redirect(routes.DetermineTaxIdController.determineTaxId())
-    }
+  val landingPage: Action[AnyContent] = as.default { implicit request =>
+    Ok(views.epayeLanding())
+  }
+
+  val landingPageSubmit: Action[AnyContent] = as.authenticatedAction { implicit request =>
```
AndyHWChung, line 40: we didn't have a `landingPageSubmit` before - what this new `val` for?

app/controllers/LandingController.scala
```diff
@@ -33,13 +33,12 @@ class LandingController @Inject() (
...
   with Logging {
 
-  val landingPage: Action[AnyContent] = as.landingPageAction { implicit request =>
-    logger.debug(s"inside landing page with journey: ${request.journey.toString}")
-    request.journey.origin match {
-      case Origins.Epaye.Bta         => Ok(views.epayeLanding())
-      case Origins.Epaye.DetachedUrl => Ok(views.epayeLanding())
-      case Origins.Epaye.GovUk       => Redirect(routes.DetermineTaxIdController.determineTaxId())
-    }
+  val landingPage: Action[AnyContent] = as.default { implicit request =>
```
AndyHWChung, line 36: do we not want to keep the old logic?

barnesjake: So at this point we don't have access to the origin - I've changed it to a default action - we may put it in somewhere else, but for now, it's not required, I don't think

AndyHWChung (approved):

PR #83, opened 2022-07-20
[JB][SUPP-376] add in ttp enact arrangement call with other stuff

barnesjake:


app/controllers/JourneyFinalStateCheck.scala
```diff
@@ -0,0 +1,57 @@
...
+  def finalStateCheckF(journey: Journey, result: => Future[Result])(implicit request: Request[_]): Future[Result] = {
+    if (endStateConditional(journey)) {
+      logMessage
+      Future.successful(confirmationRedirect)
+    } else {
+      result
+    }
+  }
+
+  def finalStateCheck(journey: Journey, result: => Result)(implicit request: Request[_]): Result = {
```
AndyHWChung, line 49: could this not be written in terms of the version above it like 
```scala
def finalStateCheck(journey: Journey, result: => Result)(implicit r: Request[_]): Result =
  finalStateCheck(journey, Future.successful(result))
```
?

app/models/forms/BankDetailsForm.scala
```diff
@@ -16,64 +16,69 @@
...
+  val accountNameConstraint: Constraint[AccountName] = {
+    Constraint(accountName =>
+      if (accountName.value.length < 100) Valid
+      else Invalid("error.invalid"))
   }
+  val accountNameMapping: Mapping[AccountName] =
+    nonEmptyText.transform[AccountName](name => AccountName(name), _.value).verifying(accountNameConstraint)
+
+  val sortCodeRegex: String = "^[0-9]{6}$"
+  val sortCodeConstraint: Constraint[SortCode] =
```
AndyHWChung, line 53: should be be stripping spaces for the sort code and account number? And maybe even the various different dashes/hyphens for sort code? That'd be more forgiving on the users which is encourages for accessibility.

barnesjake: Can do, I just copied it as it was, as I moved it out of the def

app/controllers/JourneyFinalStateCheck.scala
barnesjake: It could but it has ramifications for places it is called, i.e. chaining futures. Maybe the naming is wrong, as it's use case is pretty much take a Future[Result] and return a Future[Result]

app/models/forms/BankDetailsForm.scala
barnesjake: Oh sorry I think it does, in the sortCodeMapping:
`val sortCodeMapping: Mapping[SortCode] = nonEmptyText
    .transform[SortCode](
      s => SortCode(s.replaceAllLiterally("-", "").replaceAll("\\s", "")),
      _.value
    ).verifying(sortCodeConstraint)`

AndyHWChung: ah ok, missed that. There's a bunch of other dashes that we could filter out as well like these ones:

https://github.com/hmrc/help-to-save-frontend/blob/main/app/uk/gov/hmrc/helptosavefrontend/forms/BankDetailsValidation.scala#L43

app/controllers/JourneyFinalStateCheck.scala
AndyHWChung: oh yeah, ok - they don't both return exactly the same type. There's probably a way to make a more generic reusable version but it'd probably be overkill here I think

app/models/forms/BankDetailsForm.scala
barnesjake: I think we have a ticket to enhance the validation on the bank/direct debit details page, can we address this in that ticket?

AndyHWChung: if it's going to be revisited later on then that makes sense yeah 👍

AndyHWChung (approved):

app/models/forms/BankDetailsForm.scala
barnesjake: https://jira.tools.tax.service.gov.uk/browse/SUPP-398
It was originally for just account name, but I've altered for all fields on the screen. It does need refining though

PR #85, opened 2022-07-25
[JB][SUPP-166] add print plan page

barnesjake:


app/views/PrintSummaryPage.scala.html
```diff
@@ -63,29 +140,43 @@ <h1 class="govuk-heading-xl">
...
+        Messages.PaymentSchedule.`Monthly payments`.show
     }
-}
+    </h2>
+
+    @govukSummaryList(SummaryList(
+        rows = monthlyPaymentRows
+    ))
+
+    <p class="govuk-body">
```
nubz, line 174: we should prevent users with no JavaScript enabled from seeing this entire paragraph as it will not work for those users - we can add a custom CSS class to do that, e.g. `js-visible` 
```
<p class="govuk-body js-visible">
```

and add the necessary SASS to the application.scss file e.g. 
```
body:not(.js-enabled) .js-visible {
  display: none;
}
```
this makes use of the govuk-frontend checks that add `js-enabled` to the body when JS is detected as supported

nubz (changes requested):

app/views/PrintSummaryPage.scala.html
barnesjake: Think I've added what's required @nubz

nubz (approved):

AndyHWChung (approved):

PR #86, opened 2022-07-25
[JB][SUPP-165] missed the hmrc address details on success page

barnesjake:


app/messages/Messages.scala
```diff
@@ -683,6 +683,20 @@ object Messages {
       english = "Call the HMRC Helpline on 0300 123 1813."
```
nubz, line 683: isn't this missing `, or write to us by post:` at the end?

nubz (approved):

PR #88, opened 2022-07-27
[JB][SUPP-378/380/381/382] update ttp urls

barnesjake:
…igibility service

### DIFF
--- a/app/config/AppConfig.scala
+++ b/app/config/AppConfig.scala
@@ -35,8 +35,6 @@ class AppConfig @Inject() (config: Configuration, servicesConfig: ServicesConfig
   val authTimeoutSeconds: Int = config.get[FiniteDuration]("timeout-dialog.timeout").toSeconds.toInt
   val authTimeoutCountdownSeconds: Int = config.get[FiniteDuration]("timeout-dialog.countdown").toSeconds.toInt
 
-  val ttpBaseUrl: String = s"${servicesConfig.baseUrl("ttp")}"
-
   object BaseUrl {
     val platformHost: Option[String] = config.getOptional[String]("platform.frontend.host")
     val essttpFrontend: String = platformHost.getOrElse(config.get[String]("baseUrl.essttp-frontend"))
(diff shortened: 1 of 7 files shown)


conf/application.conf
```diff
@@ -74,6 +74,11 @@ microservice {
       host = localhost
       port = 9218
     }
```
darrell-plant, line 76: can this config `ttp` be removed?

test/testsupport/stubs/Ttp.scala
```diff
@@ -22,7 +22,7 @@ import testsupport.testdata.TtpJsonResponses
 
 object Ttp {
 
-  private val eligibilityUrl: String = "/time-to-pay/self-serve/eligibility"
+  private val eligibilityUrl: String = "/debts/time-to-pay/eligibility"
```
AndyHWChung, line 25: the vals below should be updated with the `/debts` part at the beginning as well shouldn't they?

barnesjake: They will do, but I thought I'd do it bit by bit (in the next tickets)

conf/application.conf
barnesjake: not yet, but it will in the next ticket. Eligibility has it's own microservice and different routes, annoyhingly

test/testsupport/stubs/Ttp.scala
AndyHWChung: you've just updated our own stub to have the `/debts` part in the url for all the api endpoints - the journey will be broken against our own stub if we don't update all of them here I think

AndyHWChung: it would save the hassle of raising multiple PR's if you updated all of the url's here and in the stub at once perhaps

barnesjake: Can do I don't mind

app/config/AppConfig.scala
```diff
@@ -45,6 +43,8 @@ class AppConfig @Inject() (config: Configuration, servicesConfig: ServicesConfig
     val feedbackFrontend: String = platformHost.getOrElse(config.get[String]("baseUrl.feedback-frontend"))
     val gg: String = config.get[String]("baseUrl.gg")
     val businessTaxAccountFrontend: String = platformHost.getOrElse(config.get[String]("baseUrl.business-tax-account-frontend"))
+    val timeToPayUrl: String = s"${servicesConfig.baseUrl("time-to-pay")}"
```
AndyHWChung, line 46: small thing, but `s"${x}"` is equivalent to just `x` - this and the `val` below can be cleaned up a tiny bit

barnesjake: lol sorry forgot to do that, will update

barnesjake: Done 😄

AndyHWChung (approved):

darrell-plant:
👍

PR #93, opened 2022-08-01
[DP][SUPP-423] BARs integration

darrell-plant:
Initially calling only the BARs 'validate/bank-details' endpoint.

Still todo: add tests


app/controllers/BankDetailsController.scala
```diff
@@ -99,31 +102,47 @@ class BankDetailsController @Inject() (
...
-            bankDetailsForm.isSoleSignatory.asBoolean
-          )
-
-          journeyService.updateDirectDebitDetails(request.journeyId, directDebitDetails)
-            .map { _ =>
-              bankDetailsForm.isSoleSignatory match {
-                case IsSoleSignatoryFormValue.Yes => Redirect(routes.BankDetailsController.checkBankDetails)
-                case IsSoleSignatoryFormValue.No  => Redirect(routes.BankDetailsController.cannotSetupDirectDebitOnlinePage)
-              }
+    val validForm = BankDetailsForm.form.bindFromRequest()
```
barnesjake, line 105: I think this should be called something other than validForm, as we fold on it to check for invalid form, if that makes sense

app/messages/Messages.scala
```diff
@@ -545,6 +546,11 @@ object Messages {
       english = "Select yes if you are the account holder"
     )
 
+    // BARs
```
barnesjake, line 549: Maybe put this in a object for the bars form, this file should only contain things of type Message(english: String, welsh: Option[String])

barnesjake: ALthough I'm not that fussed on this, so feel free to ignore

app/messages/Messages.scala
```diff
@@ -545,6 +546,11 @@ object Messages {
...
+    val accountNumberNotWellFormatted: FormError = FormError("bars", "validate.accountNumberIsWellFormatted.no")
```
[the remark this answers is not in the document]
AndyHWChung: you may need to do something in the views template for this page since the href of the link in the error summary is based on the key of the error but there's won't be a field with id `bars` at the moment. 

You may have to link to a specific field depending on the bars error, e.g. `accountNumberNotWellFormatted` with the account number field, `sortCodeNotPresentOnEiscd` with the sort code field etc. 

We also need to check that the error message displays above the appropriate field entry box on the page

conf/application.conf
```diff
@@ -31,6 +31,9 @@ play.modules.enabled += "uk.gov.hmrc.play.audit.AuditModule"
 play.modules.enabled += "uk.gov.hmrc.play.bootstrap.AuthModule"
 play.modules.enabled += "uk.gov.hmrc.play.bootstrap.graphite.GraphiteMetricsModule"
 
+# Enable extra frontend filters (see frontend.conf for defaults)
+play.filters.enabled += uk.gov.hmrc.play.bootstrap.frontend.filters.SessionIdFilter
```
barnesjake, line 35: Can this line be removed if we aren'y using SessionIdFilter/it's been deleted

app/controllers/BankDetailsController.scala
AndyHWChung: if you're only using this `val` once then you may as well just not have it. If you need access to it below, you could probably just use what you've called `bankDetailsForm` below

test/controllers/BankDetailsControllerSpec.scala
```diff
@@ -241,6 +241,8 @@ class BankDetailsControllerSpec extends ItSpec {
       AuthStub.authorise()
       EssttpBackend.ChosenTypeOfBankAccount.findJourney()
       EssttpBackend.DirectDebitDetails.updateDirectDebitDetails(TdAll.journeyId)
+      Bars.Validate.validateSuccess()
```
barnesjake, line 244: Could we have a test for if Bars returns a not valid response, just checking that it doesn't redirect user to confirm direct debit details or something

app/controllers/BankDetailsController.scala
```diff
@@ -99,31 +102,47 @@ class BankDetailsController @Inject() (
...
+
+          def enterBankDetailsPageWithBarsError(error: FormError): Result = {
+            Ok(views.enterBankDetailsPage(
+              validForm.withError(error),
+              BankDetailsController.chooseTypeOfAccountUrl
+            ))
+          }
+
+        journeyService.updateDirectDebitDetails(request.journeyId, directDebitDetails)
+          .flatMap { _ =>
```
AndyHWChung, line 129: it feels like we should also be updating the BE with the BARs result so we can check on subsequent pages what the result is. We wouldn't want allow someone to type in the URL of the check bank details page on the browser and see that page sucessully if the BARs result is negative

app/controllers/BankDetailsController.scala
darrell-plant: oh yeah 🙂 agreed

test/controllers/BankDetailsControllerSpec.scala
darrell-plant: Yep. "Still todo: add tests" 🙂

app/controllers/BankDetailsController.scala
darrell-plant: Yep. I think in the next ticket when we call the verify endpoint we will need to store state/stage in the BE. I was going to address that there. What do yuo think?

app/messages/Messages.scala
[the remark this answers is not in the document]
darrell-plant: > you may need to do something in the views template for this page since the href of the link in the error summary is based on the key of the error but there's won't be a field with id `bars` at the moment.
> 
> You may have to link to a specific field depending on the bars error, e.g. `accountNumberNotWellFormatted` with the account number field, `sortCodeNotPresentOnEiscd` with the sort code field etc.
> 
> We also need to check that the error message displays above the appropriate field entry box on the page

I discussed with Karen about initially not linking the errors to actual fields as with these errors it is not known which field (sortCode or accountNumber) has the error. So by using field `bars` we end up with just the error summary and an empty href.

With `accountNumberNotWellFormatted` it may be the sortCode that is wrong and with `sortCodeNotPresentOnEiscd` it may be that the accountNumber is wrong. 

So this is deliberate, and a first pass. Not sure if it is correct behaviour though. Maybe we can chat with Karen?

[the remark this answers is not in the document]
darrell-plant: > these aren't actual messages per se so it doesn't feel like they belong here

I'll think about where they can go...

app/messages/Messages.scala
darrell-plant: I'll think about where they can go...

app/models/bars/BarsModel.scala
```diff
@@ -0,0 +1,115 @@
...
+    case object Partial extends BarsAssessmentType
+  }
+
+  final case class BarsResponse(
+      accountNumberIsWellFormatted:             BarsAssessmentType,
+      nonStandardAccountDetailsRequiredForBacs: BarsAssessmentType,
+      sortCodeIsPresentOnEISCD:                 BarsAssessmentType,
+      sortCodeSupportsDirectDebit:              Option[BarsAssessmentType],
+      sortCodeSupportsDirectCredit:             Option[BarsAssessmentType],
+      sortCodeBankName:                         Option[String],
```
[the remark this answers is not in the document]
darrell-plant: Could be. We probably have one somewhere...

app/messages/Messages.scala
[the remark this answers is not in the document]
AndyHWChung: the error summary message needs to point to somewhere according to the design system here:
https://design-system.service.gov.uk/components/error-summary/
I think part of that may be for accessibility reasons so users can navigate to the relevant area quickly if they're using a screen reader. Where it's uncertain I remember linking to the first relevant box that could be wrong, so on a date if you don't know what was wrong then you'd point to the day field I think. Worth a double check, yeah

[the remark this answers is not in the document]
AndyHWChung: actually it does say this on that page so what I said about linking to the first thing that could be wrong is correct I think:
```
If you do not know which field contains an error, link to the first field.
```

app/models/bars/BarsModel.scala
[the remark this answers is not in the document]
darrell-plant: Ah, no, this is not SortCode, this is bank name

test/testsupport/stubs/Bars.scala
```diff
@@ -18,31 +18,53 @@ package testsupport.stubs
...
+      stubFor(
+        post(urlPathEqualTo(validateUrl))
+          .willReturn(
+            aResponse()
+              .withStatus(HttpStatus.SC_OK)
+              .withBody(BarsJsonResponses.validateSuccessJson)
+          )
+      )
 
+    def accountNumberNotWellFormatted(): StubMapping =
```
barnesjake, line 39: Could you simplify these to a reused method:

`def barsErrorStubMapping(responseBody: String): StubMapping = stubFor(
        post(urlPathEqualTo(validateUrl))
          .willReturn(
            aResponse()
              .withStatus(HttpStatus.SC_OK)
              .withBody(responseBody)
          )
      )`
And then each 4 of the stub mappings will be along the lines of
`def accountNumberNotWellFormatted(): StubMapping = barsErrorStubMapping(BarsJsonResponses.accountNumberNotWellFormattedJson)` etc

darrell-plant: I love refactoring. No problem 🙂

barnesjake (approved):

AndyHWChung (approved):

PR #94, opened 2022-08-02
[JB][SUPP-295] add optional text field to firestarter page to enable …

barnesjake:
…us to set empref


darrell-plant (approved):

darrell-plant:
Are you aware of `.withFormField(...)` which can be used to auto-populate fields and generate error messages? Not sure if appropriate, but see here in case it is useful:
https://confluence.tools.tax.service.gov.uk/display/TEC/2021/07/13/Help+with+date+inputs+and+error+summaries

barnesjake:
That's interesting to know, we've been using plays .fill method within controllers so far, might be worth looking into this

PR #96, opened 2022-08-03
[DP][SUPP-423] left-pad account number for BARs

darrell-plant:


app/models/bars/BarsModel.scala
```diff
@@ -39,8 +40,14 @@ object BarsModel {
 
     def apply(bankDetails: BankDetails): BarsValidateRequest =
       BarsValidateRequest(
-        BankAccount(bankDetails.sortCode, bankDetails.accountNumber)
+        BankAccount(bankDetails.sortCode, leftPad(bankDetails.accountNumber))
       )
+
+    private def leftPad(accountNumber: AccountNumber): AccountNumber = {
+      val minimumLength = 8
```
AndyHWChung, line 47: tiny optimisation here, but since this and the `val` below are constants, they can be taken outside of the def 
 so they can be created and stored once and prevent new variables being created every time this `def` is called.

darrell-plant: I can see I'm going to have to be more careful!

AndyHWChung (dismissed):

AndyHWChung (approved):

PR #97, opened 2022-08-04
[JB][SUPP-432]

barnesjake:
- add some tests
- configure what files/packages should be ignored by scoverage
- delete some unused stuff


test/controllers/DetermineTaxIdControllerSpec.scala
```diff
@@ -170,7 +170,43 @@ class DetermineTaxIdControllerSpec extends ItSpec {
       }
 
     }
+  }
+
+  "AuthenticatedActionRefiner bonus tests" - {
```
darrell-plant, line 175: bonus tests?

project/ScoverageSettings.scala
```diff
@@ -0,0 +1,27 @@
+import sbt.Def
+import scoverage.ScoverageKeys
+
+object ScoverageSettings {
+  lazy val scoverageSettings: Seq[Def.SettingsDefinition] = {
+    val excludedFiles = Seq(
+      """<empty>""",
+      """Reverse.*""",
```
AndyHWChung, line 8: would this not be covered already if we exclude the `*.Reverse.*` package below?

barnesjake: Good point, I'll check

test/controllers/DetermineTaxIdControllerSpec.scala
barnesjake: Haha sorry, poorly named but I thought I'd sneak into this class as that's the controller being used to test - I didn't want to create a whole new spec just for a few tests to cover off AuthenticatedActionRefiner

darrell-plant: Maybe better to have an `AuthenticatedActionRefinerSpec` though (some time)

barnesjake: Sure, I'll add now

darrell-plant: I'm not sure if this is a good example, but ... https://github.com/hmrc/vmv-frontend/blob/a53755bd9ec661597d615ed59bd587f2fb97f714/test/uk/gov/hmrc/vmv/actions/CcaAccountActionRefinerSpec.scala

barnesjake: It does look like it, but Pawel is _very_ against mockito 😂

AndyHWChung:
doing a spec for this would be a quick win:
image
it's a relatively simple controller endpoint where you just test the content

barnesjake:
Sure, will add now

test/controllers/DetermineTaxIdControllerSpec.scala
darrell-plant: is he against unit tests 🙂 ?

AndyHWChung (approved):

PR #106, opened 2022-08-16
[JB][SUPP-317] add audit event for Payment Plan details before submis…

barnesjake:
…sion


app/controllers/PaymentScheduleController.scala
```diff
@@ -65,9 +66,13 @@ class PaymentScheduleController @Inject() (
...
 
       case j: Journey.AfterSelectedPaymentPlan =>
-        journeyService.updateHasCheckedPaymentPlan(j.journeyId)
-          .map(_ => Redirect(routes.BankDetailsController.typeOfAccount))
-
+        for {
+          _ <- journeyService.updateHasCheckedPaymentPlan(j.journeyId)
+          _ = j match {
+            case j1: Journey.Stages.ChosenPaymentPlan => auditService.auditPaymentPlanBeforeSubmission(j1)
+            case _                                    => JourneyLogger.info("Nothing to audit, so I won't")
```
AndyHWChung, line 73: probably don't need this log at info level - maybe debug with a mention of what stage it is at so we've got the info if we need it?

app/controllers/PaymentScheduleController.scala
```diff
@@ -65,9 +66,13 @@ class PaymentScheduleController @Inject() (
         logErrorAndRouteToDefaultPageF(j)
 
       case j: Journey.AfterSelectedPaymentPlan =>
-        journeyService.updateHasCheckedPaymentPlan(j.journeyId)
-          .map(_ => Redirect(routes.BankDetailsController.typeOfAccount))
-
+        for {
+          _ <- journeyService.updateHasCheckedPaymentPlan(j.journeyId)
+          _ = j match {
```
AndyHWChung, line 71: small thing here but could we audit first and then update mongo? With audit events it's best practice to stick the event as close as possible to the thing that triggered it. Also, in the very unlikely scenario that the BE update failed then we still have a chance of having the audit event there.

barnesjake: Yeah sure thing

app/controllers/PaymentScheduleController.scala
barnesjake: Will do

AndyHWChung (approved):

PR #112, opened 2022-08-18
[JB][SUPP-315] add audit event for when we call ttp enact api

barnesjake:


app/controllers/SubmitArrangementController.scala
```diff
@@ -48,16 +52,26 @@ class SubmitArrangementController @Inject() (
...
-        .recover {
-          case e: Exception => Errors.throwServerErrorException(e.getMessage)
-        }
-    } yield Redirect(routes.PaymentPlanSetUpController.paymentPlanSetUp())
+  private def submitArrangementAndUpdateJourney(
+      journey: Journey.Stages.AgreedTermsAndConditions
+  )(implicit request: AuthenticatedJourneyRequest[_]): Future[Result] = {
+    ttpService.submitArrangement(journey)
+      .map { (r: ArrangementResponse) =>
+        auditService.auditPaymentPlanSetUp(journey, Right(r))
```
AndyHWChung, line 60: could we perhaps move the audit call to the service layer? It would be closer to the actual event there, plus if we ever need to call the service endpoint from another place then the auditing will already be taken care of. Not that that is very likely to happen though...

barnesjake: sure

app/controllers/SubmitArrangementController.scala
```diff
@@ -48,16 +52,26 @@ class SubmitArrangementController @Inject() (
...
+        r
+      }.recover {
+        case aHttpError: HttpException =>
```
darrell-plant, line 63: I'm struggling to get past this name `aHttpError` 🙂

app/models/audit/Schedule.scala
```diff
@@ -0,0 +1,62 @@
...
+    totalPaymentWithoutInterest:    AmountInPence
+)
+
+object Schedule {
+  implicit val writes: OWrites[Schedule] = Json.writes
+
+  def createSchedule(selectedPaymentPlan: PaymentPlan, dayOfMonth: DayOfMonth): Schedule = {
+    val totalNumberOfPaymentsIncludingUpfrontPayment: Int = {
+      selectedPaymentPlan.collections.regularCollections.size + {
+        if (selectedPaymentPlan.collections.initialCollection.isDefined) 1 else 0
```
darrell-plant, line 41: looks like a `fold` might be neater?

app/controllers/SubmitArrangementController.scala
barnesjake: haha, I'll update

app/models/audit/Schedule.scala
barnesjake: sure thing

app/services/AuditService.scala
```diff
@@ -190,6 +177,26 @@ class AuditService @Inject() (auditConnector: AuditConnector)(implicit ec: Execu
     )
   }
 
+  private def toPaymentPlanSetupAuditDetail(
+      journey:         AgreedTermsAndConditions,
+      responseFromTtp: Either[HttpException, ArrangementResponse]
+  )(implicit r: AuthenticatedJourneyRequest[_]): PaymentPlanSetUpAuditDetail = {
+    val maybeArrangementResponse: Option[ArrangementResponse] = responseFromTtp.toOption
+    val status: Int = responseFromTtp.fold(something => something.responseCode, _ => Status.ACCEPTED)
```
AndyHWChung, line 185: there's probably a better name than `something` here 😆 you probably could just do `_.responseCode` though and not need a name

barnesjake: haha woops 😂

darrell-plant (dismissed):

app/services/TtpService.scala
```diff
@@ -155,6 +166,19 @@ class TtpService @Inject() (ttpConnector: TtpConnector, datesService: DatesServi
     )
 
     ttpConnector.callArrangementApi(arrangementRequest, journey.correlationId)
+      .map { response: ArrangementResponse =>
+        auditService.auditPaymentPlanSetUp(journey, Right(response))
+        response
+      }.recover {
+        case httpException: HttpException =>
+          try {
```
AndyHWChung, line 174: we probably don't need a try around this audit - if we want to we could stick a try block around the `audit` method in the `AuditService` maybe. That way it will catch any audit failures - but then at that point I wouldn't necessarily throw an error - I'd just log it at warn or error level. If we can't audit then it shouldn't affect the user journey I don't think.

app/models/audit/Schedule.scala
```diff
@@ -36,11 +36,8 @@ object Schedule {
   implicit val writes: OWrites[Schedule] = Json.writes
 
   def createSchedule(selectedPaymentPlan: PaymentPlan, dayOfMonth: DayOfMonth): Schedule = {
-    val totalNumberOfPaymentsIncludingUpfrontPayment: Int = {
-      selectedPaymentPlan.collections.regularCollections.size + {
-        if (selectedPaymentPlan.collections.initialCollection.isDefined) 1 else 0
-      }
-    }
+    val totalNumberOfPaymentsIncludingUpfrontPayment: Int = selectedPaymentPlan.collections.initialCollection
+      .fold(selectedPaymentPlan.collections.regularCollections.length)(_ => selectedPaymentPlan.collections.regularCollections.length + 1)
```
darrell-plant, line 40: I was thinking
```
val totalNumberOfPaymentsIncludingUpfrontPayment: Int = 
  selectedPaymentPlan.collections.regularCollections.size + 
    selectedPaymentPlan.collections.initialCollection.fold(0)(_ => 1)
```

barnesjake: oh yeah that's nicer

darrell-plant (approved):

PR #116, opened 2022-08-23
[JB][SUPP-464] filter out any instalments that are less than £1 from …

barnesjake:
…being displayed on the instalments screen

### DIFF
--- a/test/controllers/InstalmentsControllerSpec.scala
+++ b/test/controllers/InstalmentsControllerSpec.scala
@@ -16,6 +16,9 @@
 
 package controllers
 
+import essttp.rootmodel.AmountInPence
+import essttp.rootmodel.ttp.affordablequotes.{AmountDue, PaymentPlan}
+import models.InstalmentOption
 import org.jsoup.Jsoup
 import org.jsoup.nodes.Document
 import play.api.http.Status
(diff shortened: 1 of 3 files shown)


test/controllers/InstalmentsControllerSpec.scala
```diff
@@ -130,4 +133,19 @@ class InstalmentsControllerSpec extends ItSpec {
...
   }
 
+  "InstalmentsController.retrieveInstalmentOptions" - {
+    "should filter out any instalments that are less than £1" in {
+      val plansIncludingOneLessThanAPound: List[PaymentPlan] = List(
+        TdAll.paymentPlan(1, amountDue = AmountDue(AmountInPence(1))),
+        TdAll.paymentPlan(2, amountDue = AmountDue(AmountInPence(99))),
+        TdAll.paymentPlan(3),
+        TdAll.paymentPlan(4),
+        TdAll.paymentPlan(5),
```
darrell-plant, line 143: maybe clearer to make the `amountDue` values explicit here

darrell-plant (dismissed):

darrell-plant:
So, is it okay just to filter out the smaller amounts? Does that not introduce some inaccuracy smewhere?

test/controllers/InstalmentsControllerSpec.scala
barnesjake: Sure

barnesjake:
This is only on the displaying of the instalment options. In theory if ttp only returned options less than £1 it would be an empty list, but I don't think this is possible due to validation prior to this and also ttp rules (something to do with minimum debt amount in the eligibility checks)

darrell-plant (approved):

PR #118, opened 2022-08-25
Added serviceId parameter to application.conf

tp10:


conf/application.conf
```diff
@@ -182,6 +182,6 @@ google-analytics {
   token = N/A
   host = auto
 }
-
+contact-frontend.serviceId = "essttp-frontend"
```
darrell-plant, line 185: It's probably correct, but in that file we mainly use this format
```
contact-frontend {
   serviceId = "essttp-frontend"
}
```

and I think it is better to leave a blank line before and after for readability

darrell-plant: not sure if the quotes are required either

barnesjake: I agree

PR #122, opened 2022-08-31
[JB][SUPP-471] make tax year displayed on your bill page use tax month

barnesjake:


app/controllers/YourBillController.scala
```diff
@@ -82,10 +82,13 @@ object YourBillController {
...
     val endDate: LocalDate = parseLocalDate(ass.taxPeriodTo.value)
-    InvoicePeriod(monthNumber(startDate), startDate, endDate, dueDate)
+    InvoicePeriod(monthNumberInTaxYear(startDate), startDate, endDate, dueDate)
   }
 
-  def monthNumber(date: LocalDate): Int = (date.getMonth.getValue - 4) % 12
+  def monthNumberInTaxYear(date: LocalDate): Int = {
+    val month: Int = date.getMonth.getValue
+    if (month >= 4) month - 3 else month + 9
+  }
```
darrell-plant, line 91: So the date is not important? 
e.g. monthNumberInTaxYear(1 April) is same as monthNumberInTaxYear(10 April) ?

barnesjake: Hmm that's a good point, it's for this bit 
Screenshot 2022-08-31 at 15 19 37

barnesjake: I'll add something for this, not sure if it will ever be a problem, but worth adding just in case

darrell-plant (approved):

PR #128, opened 2022-09-01
[dpp]/[SUPP-446] implement lockout after too many BARs calls

darrell-plant:
also removed the 'call both verify endpoints' logic

### DIFF
--- a/app/actions/BarsLockoutActionFilter.scala
+++ b/app/actions/BarsLockoutActionFilter.scala
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2022 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package actions
+
+import actionsmodel.AuthenticatedJourneyRequest
+import controllers.{JourneyIncorrectStateRouter, routes}
+import essttp.bars.BarsVerifyStatusConnector
+import essttp.bars.model.BarsGetStatusParams
+import essttp.journey.model.Journey
+import play.api.Logging
+import play.api.mvc.{ActionFilter, Request, Result, Results}
+
+import java.time.{Instant, ZoneId}
+import java.time.format.DateTimeFormatter
+import javax.inject.{Inject, Singleton}
+import scala.concurrent.{ExecutionContext, Future}
+
+@Singleton
+class BarsLockoutActionFilter @Inject() (barsVerifyStatusConnector: BarsVerifyStatusConnector)(
+    implicit
+    ec: ExecutionContext
+) extends ActionFilter[AuthenticatedJourneyRequest] with Logging with Results {
+
+  override protected def filter[A](journeyRequest: AuthenticatedJourneyRequest[A]): Future[Option[Result]] = {
+    implicit val rh: Request[A] = journeyRequest.request
+
+    journeyRequest.journey match {
+      case j: Journey.BeforeComputedTaxId =>
+        JourneyIncorrectStateRouter.logErrorAndRouteToDefaultPageF(j).map(Some(_))
+      case j: Journey.AfterComputedTaxId =>
+        barsVerifyStatusConnector.status(BarsGetStatusParams(j.taxId)).map { status =>
+          status.lockoutExpiryDateTime.map { expiresAt =>
+            logger.info(s"BARs lockout for taxId: ${j.taxId}")
+            Redirect(routes.BankDetailsController.barsLockout(formatted(expiresAt)))
+          }
+        }
+    }
+  }
+
+  private val fmt: DateTimeFormatter = DateTimeFormatter.ofPattern("dd MMMM yyyy, h:mma")
+
+  private def formatted(expiresAt: Instant): String = {
+    fmt.format(expiresAt.atZone(ZoneId.of("Europe/London")))
+      .replaceAll("AM$", "am")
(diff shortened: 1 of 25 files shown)


app/actions/BarsLockoutActionFilter.scala
```diff
@@ -0,0 +1,53 @@
...
+  override protected def filter[A](journeyRequest: AuthenticatedJourneyRequest[A]): Future[Option[Result]] = {
+    implicit val rh: Request[A] = journeyRequest.request
+
+    journeyRequest.journey match {
+      case j: Journey.BeforeComputedTaxId =>
+        JourneyIncorrectStateRouter.logErrorAndRouteToDefaultPageF(j).map(Some(_))
+      case j: Journey.AfterComputedTaxId =>
+        barsVerifyStatusConnector.status(j.taxId).map { status =>
+          status.lockoutExpiryDateTime.map { expiresAt =>
+            logger.info(s"BARs lockout for taxId: ${j.taxId}")
```
barnesjake, line 45: We may want to change the log level, will this log tax ids in prod?

darrell-plant: Yep. @AndyHWChung Not sure if this is an issue though. Happy to take out. We may be expected to audit this soon in any case.

app/services/EssttpBarsService.scala
```diff
@@ -42,14 +49,30 @@ class EssttpBarsService @Inject() (barsService: BarsService, auditService: Audit
   )(implicit requestHeader: RequestHeader): Future[Either[BarsError, VerifyResponse]] = {
     implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequest(requestHeader)
 
+    val taxId = journey match {
+      case account: Stages.ChosenTypeOfBankAccount     => account.taxId
```
barnesjake, line 53: can you get away with `Journey.AfterChosenTypeOfBankAccount` or similar here maybe? to avoid duplicates

AndyHWChung, line 53: you could shorten this by using `BeforeComputeTaxId` and `AfterComputeTaxId` like this perhaps:
```scala
journey match {
  case j: AfterComputeTaxId => j.taxId
  case _: BeforeComputeTaxId => throw SomeError("expected tax id but blah")
}
```

app/actions/BarsLockoutActionFilter.scala
barnesjake: I don't think we're allowed to put any taxids into kibana

app/services/EssttpBarsService.scala
darrell-plant: not sure, I can check though. IntelliJ automatically generated those case statements

darrell-plant: I'll try that

app/services/EssttpBarsService.scala
```diff
@@ -42,14 +49,30 @@ class EssttpBarsService @Inject() (barsService: BarsService, auditService: Audit
...
       subject           = toBarsSubject(bankDetails),
       business          = toBarsBusiness(bankDetails),
       typeOfBankAccount = toBarsTypeOfBankAccount(typeOfBankAccount)
-    ).map { result =>
-        auditService.auditBarsCheck(journey, bankDetails, typeOfBankAccount, result)
-        result
+    ).flatMap { result: Either[BarsError, VerifyResponse] =>
+        barsVerifyStatusConnector.update(taxId)
+          .map { verifyStatus =>
+            auditService.auditBarsCheck(journey, bankDetails, typeOfBankAccount, result)
```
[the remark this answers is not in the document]
darrell-plant: yep - one reason I left it there is that the audit may need that response from the BE update to record number of attempts? But, I can move it

app/services/EssttpBarsService.scala
```diff
@@ -50,22 +50,22 @@ class EssttpBarsService @Inject() (
...
 
     val taxId = journey match {
-      case account: Stages.ChosenTypeOfBankAccount     => account.taxId
-      case details: Stages.EnteredDirectDebitDetails   => details.taxId
-      case details: Stages.ConfirmedDirectDebitDetails => details.taxId
-      case conditions: Stages.AgreedTermsAndConditions => conditions.taxId
-      case arrangement: Stages.SubmittedArrangement    => arrangement.taxId
+      case j: AfterComputedTaxId => j.taxId
+      case _                     => throw new RuntimeException("Expected to find a taxId but none found")
     }
```
barnesjake, line 55: Sorry can this be
```
case _: BeforeComputedTaxId  => throw new RuntimeException("Expected to find a taxId but none found")
case j: AfterComputedTaxId => j.taxId
```
just incase someone ever forgets to use the marking traits on new stages

darrell-plant: doesn't compile (fruiteless type test)

app/actions/BarsLockoutActionFilter.scala
AndyHWChung: tax id's themselves are tolerated in kibana (see [here](https://confluence.tools.tax.service.gov.uk/x/D4RxHQ)) so I think this would be safe if we really wanted it

app/services/EssttpBarsService.scala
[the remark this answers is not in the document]
AndyHWChung: ah ok - yeah, when we come to do that we may need to move it. We could just move it later though when we come to it cos we don't know for certain what's gonna happen

app/controllers/BankDetailsController.scala
```diff
@@ -286,7 +288,10 @@ class BankDetailsController @Inject() (
 
   def barsLockout(p: String): Action[AnyContent] = as.default { implicit request =>
     val charset = Charset.forName("UTF-8")
-    val expiry = new String(java.util.Base64.getDecoder.decode(p.getBytes(charset)), charset)
+    val expiry = Try[String] {
+      new String(java.util.Base64.getDecoder.decode(p.getBytes(charset)), charset)
+    }.getOrElse(Instant.now.longFormat)
```
AndyHWChung, line 293: I think it's preferable to error out if `p` cannot be decoded - we always expect it to be so and providing the time now could be confusing and hide the error from us in case something did go wrong and it's our fault and not the user playing around with the parameter

AndyHWChung (approved):

PR #130, opened 2022-09-05
[DP][SUPP-446] do not update the verify count on validate errors

darrell-plant:
I missed this initially. We shouldn't be updating the number of failed calls to BARs when a validate call fails.


app/services/EssttpBarsService.scala
```diff
@@ -63,16 +63,21 @@ class EssttpBarsService @Inject() (
...
-        barsVerifyStatusConnector
-          .update(taxId)
-          .map { verifyStatus =>
-            // here we catch a lockout BarsStatus condition,
-            // and force a TooManyAttempts (BarsError) response
-            verifyStatus.lockoutExpiryDateTime
-              .fold(result) { expiry =>
-                result.flatMap(resp => Left(TooManyAttempts(resp, expiry)))
+        result match {
+          case Left(_: BarsValidateError) | Left(_: SortCodeOnDenyListValidateError) =>
```
AndyHWChung, line 67: is there a reason that `SortCodeOnDenyListValidateError` isn't a `BarsValidateError`?

app/services/EssttpBarsService.scala
```diff
@@ -63,16 +63,21 @@ class EssttpBarsService @Inject() (
...
+            Future.successful(result) // don't update the verify count on validate errors
```
AndyHWChung, line 68: are there any unit tests to test the new behaviour?

darrell-plant: There will be: https://jira.tools.tax.service.gov.uk/browse/SUPP-483

app/services/EssttpBarsService.scala
darrell-plant: the main reason is this 
```
sealed trait BarsValidateError extends BarsError {
  override val barsResponse: ValidateResponse
}

sealed trait SortCodeOnDenyListValidateError extends BarsError {
  override val barsResponse: SortCodeOnDenyList
}
```
So we can differentiate easily between different types of error response and because a ValidateResponse is a different shape than a SortCodeOnDenyList

darrell-plant: I will try this 
```sealed trait BarsError {
  val barsResponse: BarsResponse
}
sealed trait BarsValidateError extends BarsError {
  override val barsResponse: BarsResponse
}
sealed trait SortCodeOnDenyListValidateError extends BarsValidateError {
  override val barsResponse: BarsResponse
}
sealed trait BarsVerifyError extends BarsError {
  override val barsResponse: VerifyResponse
}
```

darrell-plant: updated

AndyHWChung (approved):

AndyHWChung (approved):

PR #140, opened 2022-09-12
[DP][SUPP-520] controller tests tech debt

darrell-plant:
Refactor Action stubbing (auth and bars lockout)
add a new Bars test BankDetailsController spec

### DIFF
--- a/test/testsupport/ItSpec.scala
+++ b/test/testsupport/ItSpec.scala
@@ -85,16 +85,8 @@ class ItSpec
     webDriver.manage().deleteAllCookies()
   }
 
-  // suitable for most tests
-  def stubActionDefaults(): StubMapping = {
-    stubCommonActionsWith(
-      authAllEnrolments = Some(Set(TdAll.payeEnrolment)),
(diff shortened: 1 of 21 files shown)


test/testsupport/ItSpec.scala
```diff
@@ -76,10 +82,31 @@ class ItSpec
...
+  // suitable for most tests
+  def stubActionDefaults(): StubMapping = {
+    stubCommonActionsWith(
+      authAllEnrolments = Some(Set(TdAll.payeEnrolment)),
+      authCredentials   = Some(Credentials("authId-999", "GovernmentGateway")),
+      barsLockoutExpiry = None
+    )
+  }
+  // custom action setup
+  def stubCommonActionsWith(
```
AndyHWChung, line 97: with this we may not need `stubActionDefaults` - that method is just using these default values anyway. One method it easier to maintain than two. In the one method world maybe this becomes something like `stubCommonActions` and for most cases you'd pass in nothing and use the defaults

AndyHWChung (approved):

PR #143, opened 2022-09-15
[AC][SUPP-229] Missing information page

AndyHWChung:


app/controllers/MissingInfoController.scala
```diff
@@ -0,0 +1,43 @@
...
+import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendController
+import util.Logging
+
+import javax.inject.{Inject, Singleton}
+
+@Singleton
+class MissingInfoController @Inject() (
+    as:              Actions,
+    mcc:             MessagesControllerComponents,
+    missingInfoPage: views.html.MissingInformation
```
barnesjake, line 30: Should this go into views and then use `views.missingInfoPage`. I'm not overly fussed if it doesn't but I think that's the pattern we have atm

barnesjake: By views I mean the Views class we inject

app/controllers/PaymentScheduleController.scala
```diff
@@ -45,25 +44,29 @@ class PaymentScheduleController @Inject() (
...
+        finalStateCheck(j, Ok(paymentSchedulePage(upfrontPaymentAnswers, paymentDay, j.selectedPaymentPlan)))
     }
   }
 
   val checkPaymentScheduleSubmit: Action[AnyContent] = as.eligibleJourneyAction.async { implicit request =>
     request.journey match {
-      case j: Journey.BeforeSelectedPaymentPlan =>
-        logErrorAndRouteToDefaultPageF(j)
+      case _: Journey.BeforeSelectedPaymentPlan =>
+        JourneyLogger.warn("Not enough information in session to show the check payment schedule page. " +
```
barnesjake, line 67: Could we extract this warn into a def in JourneyLogger and just pass in the page as a string maybe, saves having the same thing over and over

app/controllers/MissingInfoController.scala
AndyHWChung: 👍 I'll do that now

app/controllers/PaymentScheduleController.scala
AndyHWChung: just pushed up a change for this - it's quite specific so I wasn't sure if `JourneyLogger` was the best place to factor it out. I stuck something in the companion object of `MissingInfoController`. What do you think?

app/controllers/MissingInfoController.scala
AndyHWChung: just changed it in [this commit ](https://github.com/hmrc/essttp-frontend/pull/143/commits/85d367c7ae53ca5b0985bd793ceaf18bb235bad9)

app/controllers/PaymentScheduleController.scala
barnesjake: Yeah that sounds sensible

barnesjake (approved):

PR #161, opened 2022-09-26
[AC][SUPP-506] update bars audit event

AndyHWChung:


app/models/audit/bars/BarsVerifyDetails.scala
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.audit.bars
+
+import play.api.libs.json.{Json, OWrites}
+
+final case class BarsVerifyDetails(attempts: Int, lockoutExpiryDateTime: Option[String])
```
darrell-plant, line 21: for consistency?
```scala
attempts: NumberOfBarsVerifyAttempts,
```

darrell-plant (approved):

app/models/audit/bars/BarsVerifyDetails.scala
AndyHWChung: I've tried to keep everything to the most basic types in the audit events so it's clearer how things will get written into the JSON - with custom types there's sometimes custom serialisers which aren't what has been agreed with the CIP team. It wouldn't be the case here though cos that type would write like a simple JsNumber

darrell-plant: I was just about to delete my comment

PR #190, opened 2022-10-12
[AC][SUPP-575] radio checked on error; cleanup

AndyHWChung:


barnesjake (approved):

app/actions/AuthenticatedActionRefiner.scala
```diff
@@ -17,44 +17,41 @@
...
-    appConfig: AppConfig,
-    cc:        MessagesControllerComponents
+    val authConnector: AuthConnector,
+    appConfig:         AppConfig,
+    cc:                MessagesControllerComponents
 )(
     implicit
     ec: ExecutionContext
-) extends ActionRefiner[Request, AuthenticatedRequest] {
+) extends ActionRefiner[Request, AuthenticatedRequest] with AuthorisedFunctions {
```
pawelpanasewicz, line 41: It was intentionally coded that way to promote composition over inheritance. It seemed to me more clear.

```
af.authorised()
```

as opposed to 
```
authorised()
``` 
clearly says where the function comes from.

More over it's not obvious what is the role of `val authConnector: AuthConnector,` on the constructor list.

AndyHWChung: ah ok, it just requires a bit more set up in binding `AuthorisedFunctions` in a module. This way was suggested in during our PRA - it seems cleaner at least with this platform standard

PR #202, opened 2022-10-25
[JB][SUPP-495] add vat to firestarter page

barnesjake:


app/testOnly/models/formsmodel/StartJourneyForm.scala
```diff
@@ -18,38 +18,41 @@ package testOnly.models.formsmodel
...
-    eligibilityErrors: Seq[EligibilityError],
-    debtTotalAmount:   BigDecimal,
-    interestAmount:    Option[BigDecimal],
-    taxReference:      Option[String]
+    signInAs:            SignInAs,
+    enrolments:          Seq[Enrolment],
+    origin:              Origin,
+    eligibilityErrors:   Seq[EligibilityError],
+    payeDebtTotalAmount: Option[BigDecimal],
+    vatDebtTotalAmount:  Option[BigDecimal],
```
AndyHWChung, line 38: instead of two `Option[BigDecimal]` for `payeDebtTotalAmount` and `vatDebtTotalAmount` it could be cleaner to have a separate `taxRegime: String` field which is a hidden field on each tab but can be used in the mapping to tell which tax regime we're in. Then you could have one non-optional `debtTotalAmount: BigDecimal` and perhaps even just one `taxReference: Option[String]`. What do you think?

barnesjake: That does make sense, but the trouble I had was trying to obtain the input tax ref and using that to drive the validation on the amount field, have you got any examples of doing this?

barnesjake: I think part of my issue is I also tried to make the amount without interest field optional (so it just defaults to £1234.53) for ease of use to save time getting into the journey

app/testOnly/models/formsmodel/StartJourneyForm.scala
```diff
@@ -67,25 +68,81 @@ object StartJourneyForm {
...
-    val debtTotalAmountMapping: FieldMapping[BigDecimal] = Forms.of(amountOfMoneyFormatter(
-      isTooSmall = AmountInPence(100) > AmountInPence(_),
-      isTooLarge = AmountInPence(_) > maxAmountOfDebt
-    ))
-
     val interestAmountMapping: Mapping[Option[BigDecimal]] =
       optional(Forms.of(amountOfMoneyFormatter(_ < 0, _ => false)))
 
-    val taxRefMapping: Mapping[Option[String]] = optional(Forms.text)
+    //    val taxRefMapping: Mapping[Option[String]] = optional(Forms.text)
```
AndyHWChung, line 74: this can be cleaned up

app/testOnly/models/formsmodel/StartJourneyForm.scala
```diff
@@ -67,25 +68,81 @@ object StartJourneyForm {
...
+          taxRegimeFormatter.bind(taxRegimeKey, data).map{ taxRegime =>
+            val (taxReferenceKey, defaultTaxRef): (String, TaxId) = taxRegime match {
+              case TaxRegime.Epaye => payeTaxReferenceKey -> RandomDataGenerator.nextEpayeRefs()(Random)._3
+              case TaxRegime.Vat   => vatTaxReferenceKey -> RandomDataGenerator.nextVrn()(Random)
+            }
+            val dataWithDefaultValue: Map[String, String] =
+              if (data.get(taxReferenceKey).exists(_.nonEmpty)) data else data.updated(taxReferenceKey, defaultTaxRef.value)
+
+            val taxRef: TaxId = taxRegime match {
+              case TaxRegime.Epaye => EmpRef(dataWithDefaultValue.getOrElse(taxReferenceKey, defaultTaxRef.value))
```
AndyHWChung, line 126: if you're doing `getOrElse` over here then you don't really need `dataWithDefaultValue` - you might as well do what you're doing here with `data`

app/testOnly/models/formsmodel/StartJourneyForm.scala
```diff
@@ -67,25 +68,81 @@ object StartJourneyForm {
...
+              case TaxRegime.Vat   => Vrn(dataWithDefaultValue.getOrElse(taxReferenceKey, defaultTaxRef.value))
+            }
+            taxRef
+          }
+
+        // we don't actually need the unbind, but because we've created a Formatter, we need to declare one...
+        override def unbind(key: String, value: TaxId): Map[String, String] = ???
```
AndyHWChung, line 133: the `unbind` is needed to transform from the case class to the underling data map in the form - this is typically used when prepopulating the data on the page. It should be easy to write this one though - you just look at the tax id type to determine if the key should be the PAYE or VAT one and create a map where that key maps onto the value of the tax id.

barnesjake: ok

app/testOnly/models/formsmodel/StartJourneyForm.scala
```diff
@@ -117,14 +117,15 @@ object StartJourneyForm {
               case TaxRegime.Epaye => payeTaxReferenceKey -> RandomDataGenerator.nextEpayeRefs()(Random)._3
               case TaxRegime.Vat   => vatTaxReferenceKey -> RandomDataGenerator.nextVrn()(Random)
             }
-            val dataWithDefaultValue: Map[String, String] =
-              if (data.get(taxReferenceKey).exists(_.nonEmpty)) data else data.updated(taxReferenceKey, defaultTaxRef.value)
-
-            val taxRef: TaxId = taxRegime match {
-              case TaxRegime.Epaye => EmpRef(dataWithDefaultValue.getOrElse(taxReferenceKey, defaultTaxRef.value))
-              case TaxRegime.Vat   => Vrn(dataWithDefaultValue.getOrElse(taxReferenceKey, defaultTaxRef.value))
+            val dataWithDefaultValue: String = {
```
AndyHWChung, line 120: couldn't you do something like 
```scala
data.get(taxReferenceKey)
  .filter(_.nonEmpty)
  .map(s => taxRegimeMatch {
    case TaxRegime.Epaye => EmpRef(s)
    case TaxRegime.Vat => Vrn(s)
  })
  .getOrElse(defaultTaxRef)
```
then you don't need to update a map with a default value only then to do a `getOrElse` on it anyway

barnesjake: will do

AndyHWChung (approved):

PR #203, opened 2022-10-27
Jake/supp 493

barnesjake:


app/messages/Messages.scala
```diff
@@ -404,6 +404,71 @@ object Messages {
...
+      english = "The payment plan covers all appropriate overdue amounts, surcharges, penalties and interest. The payments you make may incur interest."
+    )
+    val `Who can use this service`: Message = Message(
+      english = "Who can use this service"
+    )
+    val `You are eligible to set up an online payment plan if:`: Message = Message(
+      english = "You are eligible to set up an online payment plan if:",
+      welsh   = "Rydych yn gymwys i drefnu cynllun talu ar-lein os:"
+    )
+    def `you plan to pay the debt off within the next 6 months or less`(maxPlanDuration: Int): Message = Message(
```
AndyHWChung, line 421: small thing but `6` in the val name doesn't reflect the input to this `def` - I've put `...` in similar `val` names for EPAYE where the value is variable

app/messages/Messages.scala
```diff
@@ -404,6 +404,71 @@ object Messages {
...
+      english = s"you plan to pay the debt off within the next $maxPlanDuration months or less"
+    )
+    def `you owe £20,000 or less`(maxAmountOfDebt: AmountInPence): Message = Message(
```
AndyHWChung, line 424: same thing here

app/messages/Messages.scala
```diff
@@ -404,6 +404,71 @@ object Messages {
...
+      english = s"you owe ${maxAmountOfDebt.gdsFormatInPounds} or less"
+    )
+    val `you do not have any other payment plans or debts with HMRC`: Message = Message(
+      english = "you do not have any other payment plans or debts with HMRC"
+    )
+    val `your tax returns are up to date`: Message = Message(
+      english = "your tax returns are up to date"
+    )
+    def `You can use this service within 28 days of the overdue payment deadline.`(maxAgeOfDebtInDays: Int): Message = Message(
```
AndyHWChung, line 433: and here

app/views/templates/Layout.scala.html
```diff
@@ -51,7 +51,8 @@
   additionalScripts: Option[Html] = None,
   showBackLink: Boolean = true,
   hasErrors: Boolean = false,
-  showSignOutOverride: Option[Boolean] = None
+  showSignOutOverride: Option[Boolean] = None,
+  taxRegime: TaxRegime = TaxRegime.Epaye
```
AndyHWChung, line 55: a default argument here for `taxRegime` seems a bit dangerous - there's a way to get the taxRegime without an extra parameter here if you look at `serviceName` below

app/views/vat/VatLanding.scala.html
```diff
@@ -0,0 +1,79 @@
...
+        <li>@{Messages.Vat.`cash accounting customers`.show}</li>
+        <li>@{Messages.Vat.`annual accounting scheme members`.show}</li>
+        <li>@{Messages.Vat.`payment on account customers`.show}</li>
+    </ul>
+
+    <p class="govuk-body">@{Messages.Epaye.`You must keep up to date with your payments...`.show}</p>
+
+    @govukButton(Button(
+        href = Some(routes.DetermineTaxIdController.determineTaxId.url),
+        content = Text(Messages.Epaye.`Start now`.show)
```
AndyHWChung, line 77: `Start now` can be moved up a level out of `Epaye` I think since it's generic

app/messages/Messages.scala
barnesjake: ah ok! Will update

app/views/vat/VatLanding.scala.html
barnesjake: Ok

app/views/templates/Layout.scala.html
barnesjake: I did this to cater for when users haven't started a journey yet and navigate directly to it, and used the default to not have to change all of the Paye pages, then I thought when we introduce the "what regime do you want to pay" page we could remove it.

AndyHWChung: for that case maybe we can have a temporary message like `Set up a payment plan` - so we had this previously:
```scala
  case r: AuthenticatedJourneyRequest[_] => Messages.ServicePhase.serviceName(r.journey.taxRegime).show
  case _ => Messages.ServicePhase.serviceName(TaxRegime.Epaye).show
```
Perhaps this could be:
:
```scala
  case r: AuthenticatedJourneyRequest[_] => Messages.ServicePhase.serviceName(r.journey.taxRegime).show
  case _ => Messages.ServicePhase.`Set up a payment plan`
```
with a TODO there perhaps to double check what that default case should be

barnesjake: Ah yeah that's better to be fair, I'll implement

barnesjake:
Actually can we hold fire on this one, I want to try and get the new origin one in first as it's higher priority

AndyHWChung (approved):

PR #212, opened 2022-10-31
Jake/supp 554

barnesjake:


app/views/vat/NotVatRegistered.scala.html
```diff
@@ -0,0 +1,81 @@
...
+}
+@layout(pageTitle = Some(Messages.EnrolmentMissing.`You are not registered`.show), showBackLink = false) {
+    <h1 class="govuk-heading-xl">
+    @{Messages.EnrolmentMissing.`You are not registered`.show}
+    </h1>
+    <p class="govuk-body">
+        @{Messages.EnrolmentMissing.`You are not eligible for an online payment plan because you need to register for VAT Online.`.show}
+        <a href="@{appConfig.Urls.enrolForVatUrl}" class="govuk-link" id="how-to-enrol-link">@{Messages.EnrolmentMissing.`Find out how to register.`.show}</a>@{Messages.`full stop`.show}
+    </p>
+    <div class="common-eligibility">
```
AndyHWChung, line 54: is `common-eligibility` actually a class? Was it meant to be an id? Is it even needed?

barnesjake: oops, I just copy pasted it from the others, I'll change them all to ids

barnesjake: I think I originally added it to make the content assertions in the tests easier

AndyHWChung (approved):

PR #216, opened 2022-11-01
[JB] fix nav bar content for not enrolled pages, sign out pages, bars lockout page

barnesjake:


app/controllers/SignOutController.scala
```diff
@@ -54,21 +54,24 @@ class SignOutController @Inject() (
   }
 
   def doYouWantToGiveFeedback: Action[AnyContent] = Action { implicit request =>
-    Ok(views.doYouWantToGiveFeedbackPage(GiveFeedbackForm.form))
+    val taxRegimeString: String = request.session.get(SignOutController.feedbackRegimeKey).getOrElse("NA")
+    val maybeFallbackTaxRegime: Option[TaxRegime] = TaxRegime.withNameInsensitiveOption(taxRegimeString)
```
AndyHWChung, line 58: if we can't get a tax regime here then something has gone wrong - you could extract out what we had before perhaps into a separate method and use it here and in the submit:
```scala
val taxRegimeString = request.session.get(SignOutController.feedbackRegimeKey)
    .getOrElse(sys.error("Could not find tax regime in cookie session"))
```

barnesjake: Ok will do

app/controllers/SignOutController.scala
```diff
@@ -82,6 +80,10 @@ class SignOutController @Inject() (
   val exitSurveyPaye: Action[AnyContent] = Action { _ =>
     Redirect(appConfig.ExitSurvey.payeExitSurveyUrl).withNewSession
   }
+
+  def maybeTaxRegimeString(request: MessagesRequest[_]): String =
```
AndyHWChung, line 84: small thing, but it's not maybe anymore since it's not an option

PR #221, opened 2022-11-03
[JB][SUPP-599] add in label

barnesjake:


app/views/MonthlyPaymentAmount.scala.html
```diff
@@ -67,19 +67,19 @@
...
+        content = HtmlContent(detailsContent)
+    ))
 
-        @govukDetails(Details(
-            summary = Text(Messages.MonthlyPaymentAmount.`I can’t afford the minimum payment`.show),
-            content = HtmlContent(detailsContent)
-        ))
+    @formWithCSRF(action = routes.MonthlyPaymentAmountController.monthlyPaymentAmountSubmit) {
 
+        <label for="@key">
```
AndyHWChung, line 82: is there a way of using `label` in the parameters to `Input` here:
https://github.com/hmrc/play-frontend-hmrc/blob/main/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/input/Input.scala#L33
? Does it clash in any way with the label that's created already by the input here?
https://github.com/hmrc/play-frontend-hmrc/blob/main/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukInput.scala.html#L25

barnesjake: If I use the label param in the input, it puts some extra text in which doesn't work, can we catch up after standup?

AndyHWChung (approved):

PR #232, opened 2022-11-07
[JB][SUPP-604] add VatService origin

barnesjake:


app/testOnly/controllers/StartJourneyController.scala
```diff
@@ -158,7 +151,17 @@ class StartJourneyController @Inject() (
         returnUrl = ReturnUrl(routes.StartJourneyController.showBtaVatPage.url + "?return-page"),
         backUrl   = BackUrl(routes.StartJourneyController.showBtaVatPage.url + "?starting-page")
       )).map(sjResponse => Redirect(sjResponse.nextUrl.value))
-      //        Redirect(_root_.controllers.routes.LandingController.vatLandingPage))
+    }
+  }
+
+  val startJourneyVatVatService: Action[AnyContent] = as.default.async { implicit request =>
+    if (hc.sessionId.isEmpty) {
```
AndyHWChung, line 158: if `missingSessionOrTestPage` was something like
```scala
private def withSessionId(result: => Future[Result])(implicit request: Request[_]):Future[Result] =
    if (hc.sessionId.isEmpty) Future.successful(Ok("Missing session id")) else result
```
then you could reuse it here and in similar places as well

barnesjake: ah good point, I'll update

PR #236, opened 2022-11-08
[JB][SUPP-593] Amend arrangement/enact API models for VAT

barnesjake:


app/services/TtpService.scala
```diff
@@ -234,4 +240,21 @@ object TtpService {
     case someAmount: UpfrontPaymentAnswers.DeclaredUpfrontPayment => Some(someAmount.amount)
     case UpfrontPaymentAnswers.NoUpfrontPayment                   => None
   }
+
+  /**
+   * If email matches the one from the eligibility API - ETMP
+   * If new email entered on the screen which doesn't match the ETMP one - TEMP
+   */
+  private def deriveCustomerDetail(journey: Journey.Stages.EmailVerificationComplete): Option[List[CustomerDetail]] = {
+    val emailFromEligibilityResponse: Option[List[Option[String]]] =
```
AndyHWChung, line 249: I think we'd need to only check against emails in the eligibility response which have type `emailSource` equal to `ETMP` so we could apply a filter to the `List` in the eligibility response. 

Also, since email addresses are case insensitive, we could `.toLowerCase` the eligibility response emails and the email in our journey before checking for inequality.

app/testOnly/controllers/StartJourneyController.scala
```diff
@@ -163,13 +164,13 @@ object StartJourneyController {
     case "Individual"   => uk.gov.hmrc.auth.core.AffinityGroup.Individual
   }
 
-  private def makeEligibilityCheckResult(form: StartJourneyForm): EligibilityCheckResult = {
-
-    val debtAmountFromForm: AmountInPence =
-      AmountInPence(form.debtTotalAmount)
+  private def makeEligibilityCheckResult(form: StartJourneyForm, emailJourneyEnabled: Boolean): EligibilityCheckResult = {
```
AndyHWChung, line 167: it probably won't make any difference if we prime the stub with email info even if the email journey is disabled - it'd just get ignored anyway. The enact request should be the same as well since no emails would have been given to us by the user if the email journey is disabled.

barnesjake: True, updated

app/services/TtpService.scala
barnesjake: I think I've implemented how you've asked

app/services/TtpService.scala
```diff
@@ -234,4 +242,25 @@ object TtpService {
...
+
+  /**
+   * If email matches the one from the eligibility API - ETMP
+   * If new email entered on the screen which doesn't match the ETMP one - TEMP
+   */
+  private def deriveCustomerDetail(journey: Journey.Stages.EmailVerificationComplete): Option[List[CustomerDetail]] = {
+    val etmpEmails: Option[List[CustomerDetail]] =
+      journey.eligibilityCheckResult.customerDetails.map(_.filter(_.emailSource.contains(EmailSource.ETMP)))
+
+    val emailThatsBeenVerified: String = journey.emailVerificationAnswers match {
```
AndyHWChung, line 254: I think you can just do `journey.emailToBeVerified` here to get the email address

app/services/TtpService.scala
```diff
@@ -234,4 +242,25 @@ object TtpService {
...
+      case EmailVerificationAnswers.EmailVerified(email, _) => email.value.decryptedValue
+      case EmailVerificationAnswers.NoEmailJourney          => Errors.throwBadRequestException("There was no email when there should be.")
+    }
+
+    val maybeEtmpEmail: Option[List[CustomerDetail]] =
+      etmpEmails.map(_.filter(_.emailAddress.map(_.toLowerCase(Locale.UK)) === Some(emailThatsBeenVerified.toLowerCase(Locale.UK))))
```
AndyHWChung, line 260: `.filter` on the left doesn't seem right. Wouldn't it be something like this?
```scala
etmpEmails
  .map( _.map(_.emailAddress.map(_.toLowerCase(Locale.UK)))
  .exists(_.contains(Some(emailThatsBeenVerified.toLowerCase(Locale.UK)))
```

AndyHWChung: oh wait I'm reading it wrong, nevermind!

app/services/TtpService.scala
```diff
@@ -142,20 +146,22 @@ class TtpService @Inject() (
     val directDebitDetails = journey.fold(_.directDebitDetails, _.directDebitDetails)
     val selectedPaymentPlan = journey.fold(_.selectedPaymentPlan, _.selectedPaymentPlan)
     val correlationId = journey.fold(_.correlationId, _.correlationId)
+    val regimeDigitalCorrespondence = journey.fold(_.eligibilityCheckResult.regimeDigitalCorrespondence, _.eligibilityCheckResult.regimeDigitalCorrespondence)
+    val customerDetail = journey.fold(_.eligibilityCheckResult.customerDetails, deriveCustomerDetail)
```
AndyHWChung, line 150: since this and `regimeDigitalCorrespondence` are being added now, it'd be handy to add unit tests testing that the correct values are populated in the request - especially around the email and the email source and toLowerCase stuff

app/services/TtpService.scala
barnesjake: ok

app/services/TtpService.scala
barnesjake: sure thing

AndyHWChung (approved):

PR #244, opened 2022-11-16
[AC][SUPP-633] prevent multiple submit arrangement requests

AndyHWChung:


AndyHWChung:
`prevent-double-click `only prevents double clicks within a certain time period:
https://github.com/alphagov/govuk-frontend/issues/2498
I used the workaround mentioned there on top of `prevent-double-click` which disables the submit button on form submit to prevent multiple arrangements requests firing

barnesjake (approved):

PR #247, opened 2022-11-16
[JB][SUPP-636] enable vat journey to update upfront payment yes/no

barnesjake:


test/controllers/UpfrontPaymentControllerSpec.scala
```diff
@@ -43,121 +45,130 @@ class UpfrontPaymentControllerSpec extends ItSpec {
...
-      EssttpBackend.CanPayUpfront.findJourney(testCrypto)()
+  Seq[(String, Origin, Enrolment, TaxRegime)](
+    ("EPAYE", Origins.Epaye.Bta, TdAll.payeEnrolment, TaxRegime.Epaye),
+    ("VAT", Origins.Vat.Bta, TdAll.vatEnrolment, TaxRegime.Vat)
+  ).foreach{
+      case (regime, origin, enrolment, taxRegime) =>
+        "GET /can-you-make-an-upfront-payment" - {
+
+          s"[$regime journey] should return 200 and the can you make an upfront payment page" in {
+            stubCommonActions(Some(Set(enrolment)))
```
AndyHWChung, line 56: do the enrolments actually make a difference here? Don't the enrolments only really matter when looking for the tax id?

AndyHWChung: seems to work ok when I change this for `stubCommonActions()` without the enrolments on my local machine. Saves a bit of code if it's not needed

barnesjake: Ok will update

AndyHWChung (approved):

PR #248, opened 2022-11-17
[JB][SUPP-637/638] stages for vat to get facilitate upfront payment a…

barnesjake:
…mount

### DIFF
--- a/app/controllers/MonthlyPaymentAmountController.scala
+++ b/app/controllers/MonthlyPaymentAmountController.scala
@@ -19,12 +19,12 @@ package controllers
 import _root_.actions.Actions
 import controllers.JourneyFinalStateCheck.finalStateCheck
 import controllers.JourneyIncorrectStateRouter.logErrorAndRouteToDefaultPage
-import controllers.MonthlyPaymentAmountController.monthlyPaymentAmountForm
+import controllers.MonthlyPaymentAmountController.{monthlyPaymentAmountForm, upfrontPaymentAnswersFromJourney}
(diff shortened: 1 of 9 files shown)


app/controllers/MonthlyPaymentAmountController.scala
```diff
@@ -162,4 +161,23 @@ object MonthlyPaymentAmountController {
       key -> Forms.of(amountOfMoneyFormatter(minimumAmount.inPounds > _, maximumAmount.inPounds < _))
     )(identity)(Some(_))
   )
+
+  private def upfrontPaymentAnswersFromJourney(journey: Journey.AfterRetrievedAffordabilityResult): UpfrontPaymentAnswers = {
```
barnesjake, line 165: Sorry, this is probably the worst bit about the pr, we can't use the old way of `j.into[Journey.AfterUpfrontPaymentAnswers]` as chimney complains about ambiguity, same for the similar approach in TtpService below

AndyHWChung: we may be able to have a shorter pattern match at least here maybe - does something like
```scala
journey match {
  case j: Journey.AfterUpfrontPaymentAnswers => j.upfrontPaymentsAnswers
  case j: Journey.BeforeUpfrontPaymentAnswers => // error
```
work? Though the second case is annoying cos we don't actually have a way of reaching it atm

barnesjake: updated

AndyHWChung (approved):

PR #270, opened 2022-12-07
[JB][SUPP-679] enhancements to firestarter page, remember answers on …

barnesjake:
…form error

### DIFF
--- a/app/testOnly/views/TestOnlyStartPage.scala.html
+++ b/app/testOnly/views/TestOnlyStartPage.scala.html
@@ -48,13 +48,15 @@
 @title = @{
     "Test only starter page"
 }
+
 @eligibilityErrorCheckboxes = @{
     EligibilityErrors.values
             .filterNot(_ === EligibilityErrors.MultipleReasons)
             .map { error =>
         CheckboxItem(
             content = Text(error.entryName),
-            value = error.entryName
+            value = error.entryName,
+            checked = form.data.exists(_._2 === error.entryName)
         )
     }
 }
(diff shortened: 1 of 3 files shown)


app/testOnly/views/TestOnlyStartPage.scala.html
```diff
@@ -378,23 +412,9 @@ <h1 class="govuk-heading-xl">@title</h1>
...
-            TabItem(
-                id = Some("vat"),
-                label = "VAT",
-                panel = TabPanel(
-                    content = HtmlContent(vatContent)
-                )
-            )
-        )
+        items =
+            if (form.data.exists(taxRegime => taxRegime._1 === "taxRegime" && taxRegime._2 == "Vat")) Seq(vatTab, epayeTab)
```
AndyHWChung, line 416: might be a bit confusing with the order of the tabs suddenly changing - on the other hand there doesn't seem to be a way to change the preselected tab using play-frontend-hmrc 🤔

https://github.com/hmrc/play-frontend-hmrc/blob/main/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukTabs.scala.html#L29

I wonder if there's a javascript way we can change the preselection without changing the order of the tabs - unless we don't care about the order of the tabs changing. What do you think @barnesjake @DarrenCB ?

barnesjake: I did toy around with trying to use the css attributes to change tabIndex on load but it didn't work (I was trying to keep ordering the same but the active tab as whatever related to the form submission), happy to see if there's a js way

DarrenCB: Would be nice to keep same order of tabs, but I wouldn't say it's a huge deal if not possible.

barnesjake: It does only reorder on form submission error 🤷‍♂️ Maybe we could ask Tim?

AndyHWChung (approved):

PR #281, opened 2022-12-19
[DCB][SUPP-579] update Welsh translations

DarrenCB:


app/messages/Messages.scala
```diff
@@ -352,6 +352,11 @@ object Messages {
       welsh   = "Dysgu sut i gofrestru"
     )
 
+    val `If you need to speak to an adviser call us on 0300 200 3835 at the Business Support Service to talk about your payment options.`: Message = Message(
```
barnesjake, line 355: Where's this new content come from? I can't see it in the prototype v3.1

app/messages/Messages.scala
```diff
@@ -1318,9 +1323,9 @@ object Messages {
       welsh   = "Yr hyn sy’n digwydd nesaf"
     )
 
-    val `HMRC will send you a letter within 5 working days with your payment dates.`: Message = Message(
-      english = "HMRC will send you a letter within 5 working days with your payment dates.",
-      welsh   = "Bydd CThEF yn anfon llythyr atoch cyn pen 5 diwrnod gyda’ch dyddiadau talu."
+    val `We will send a letter by post confirming the set up of your Direct Debit instruction within 5 working days.`: Message = Message(
```
barnesjake, line 1326: Same with this

app/messages/Messages.scala
DarrenCB: Second paragraph:
https://supp-prototype.herokuapp.com/3-1/unhappy/not-enrolled
https://supp-prototype.herokuapp.com/3-1/vat/not-enrolled

app/messages/Messages.scala
DarrenCB: Confirmation page prototype:
https://supp-prototype.herokuapp.com/3-1/confirmation

Karen just confirming with business if wording needs to be applied for both EPAYE & VAT. For now, updating. If need to make them separate, can do that later.

app/messages/Messages.scala
barnesjake: I feel like it should be one message or the other, and maybe prototype isn't synced up - both messages are basically the same just different word ordering within the sentence. Also, can we get a ticket for this? I don't want to start merging in content changes without a ticket as it's technically not properly testable - we can't just say "check protoytype"

We need a ticket saying, it was "this" it should be "this" and list each page

DarrenCB: Also in Welsh spreadsheet: https://docs.google.com/spreadsheets/d/1CdBoO5ASk4iwsvHStPkT36tbIZNduLOC/edit#gid=47327513

DarrenCB: I will create a ticket & double-check exactly which wording with Rocco. But I am following the spreadsheet of content in ticket 579.

barnesjake (approved):

barnesjake (approved):

PR #283, opened 2023-01-03
[JB][SUPP-680] store email check status in backend

barnesjake:


app/controllers/EmailController.scala
```diff
@@ -112,17 +112,31 @@ class EmailController @Inject() (
                   case Some(email) => Email(SensitiveString(email))
                   case None        => Email(emailFromEligibilityResponse.value)
                 }
-                journeyService
-                  .updateSelectedEmailToBeVerified(
+                val emailVerificationResult: Option[EmailVerificationResult] = request.journey match {
```
AndyHWChung, line 115: I think we'd need to make sure that the verification result here is definitely for the same email address the user has just submitted - otherwise you could be reading a verification status for a different one. But also, I think you may not need this - the part about handling the case if it's already been verified or locked will come in the update to `requestVerification` if you leave this submit and the one just below as it was

barnesjake: Ok, I'll change back to what it was and test locally

barnesjake: So if I don't have these checks in, it gets stuck in a loop of too many redirects due to the state not being correct

app/services/EmailVerificationService.scala
```diff
@@ -48,7 +47,7 @@ class EmailVerificationService @Inject() (
     connector.startEmailVerificationJourney(emailVerificationRequest(emailAddress))
 
   def getEmailVerificationResult(emailAddress: Email)(implicit r: EligibleJourneyRequest[_], hc: HeaderCarrier): Future[EmailVerificationResult] =
-    connector.getEmailVerificationResult(GetEmailVerificationResultRequest(r.ggCredId, emailAddress))
+    connector.getEmailVerificationResult(GetEmailVerificationResultRequest(r.ggCredId, emailAddress)) //todo jake return result instead of state
```
AndyHWChung, line 50: this TODO can be cleaned up now right?

barnesjake: whoops yeah

AndyHWChung (approved):

PR #296, opened 2023-01-16
[JB][SUPP-697]

barnesjake:
- update existing audit events to include new fields
- change email-verification endpoints to pass journeyId to backend

### DIFF
--- a/app/controllers/PaymentScheduleController.scala
+++ b/app/controllers/PaymentScheduleController.scala
@@ -17,7 +17,6 @@
 package controllers
 
 import _root_.actions.Actions
-import config.AppConfig
 import controllers.JourneyFinalStateCheck.finalStateCheck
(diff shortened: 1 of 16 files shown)


app/controllers/PaymentScheduleController.scala
```diff
@@ -74,8 +76,10 @@ class PaymentScheduleController @Inject() (
 
       case j: Journey.AfterSelectedPaymentPlan =>
         j match {
```
barnesjake, line 78: This conditional is essentially so we audit for non email journeys at this point, but don't audit if it is an email journey - acceptance criteria asks for email info to be added to the event before submission (only applicable to email journeys).

barnesjake, line 78: This conditional is essentially so we audit for non email journeys at this point, but don't audit if it is an email journey - acceptance criteria asks for email info to be added to the event before submission (only applicable to email journeys).

AndyHWChung (approved):

PR #297, opened 2023-01-18
[AC][SUPP-707] remove service name from landing page headers

AndyHWChung:


barnesjake:
I'll approve, but not sure we should be removing it (Is it the content in the black nav bar at the top? If it is I would argue that SA SUPP is wrong and we're making our service worse if it is - SA SUPP should add it in, not us remove it)

barnesjake (approved):

PR #298, opened 2023-01-20
[JB][SUPP-708] update too many passcode journeys error page

barnesjake:


app/messages/Messages.scala
```diff
@@ -1174,14 +1174,19 @@ object Messages {
 
   object TooManyPasscodeJourneys {
 
-    val `You have tried to confirm an email too many times`: Message = Message(
-      english = "You have tried to confirm an email too many times",
-      welsh   = "Rydych wedi ceisio cadarnhau cyfeiriad e-bost gormod o weithiau"
+    val `You have tried to verify an email address too many times`: Message = Message(
+      english = "You have tried to verify an email address too many times",
+      welsh   = "ADD_ME"
```
AndyHWChung, line 1179: if we merge this in now then it'd be a good idea to create a placeholder ticket to add the Welsh so we don't forget

barnesjake: Good point, will raise now

barnesjake: Ticket raised: https://jira.tools.tax.service.gov.uk/browse/SUPP-709
And mentioned it in the welsh content placeholder

AndyHWChung (approved):

PR #317, opened 2023-02-27
[JB][SUPP-737] Additional Eligibility failure - charge lock code fail…

barnesjake:
…ing to match, also couple missing tests for the other optional eligibilityRules

### DIFF
--- a/test/controllers/DetermineEligibilityControllerSpec.scala
+++ b/test/controllers/DetermineEligibilityControllerSpec.scala
@@ -80,7 +80,22 @@ class DetermineEligibilityControllerSpec extends ItSpec {
         Origins.Epaye.Bta),
       ("NoDueDatesReached - VAT", TdAll.notEligibleNoDueDatesReached, "noDueDatesReached", PageUrls.vatNotEligibleUrl,
         JourneyJsonTemplates.`Eligibility Checked - Ineligible - NoDueDatesReached`(Origins.Vat.Bta),
-        Origins.Vat.Bta)
+        Origins.Vat.Bta),
+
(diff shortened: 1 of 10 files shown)


test/controllers/DetermineEligibilityControllerSpec.scala
```diff
@@ -80,7 +80,22 @@ class DetermineEligibilityControllerSpec extends ItSpec {
         Origins.Epaye.Bta),
       ("NoDueDatesReached - VAT", TdAll.notEligibleNoDueDatesReached, "noDueDatesReached", PageUrls.vatNotEligibleUrl,
         JourneyJsonTemplates.`Eligibility Checked - Ineligible - NoDueDatesReached`(Origins.Vat.Bta),
-        Origins.Vat.Bta)
+        Origins.Vat.Bta),
+
```
barnesjake, line 84: we didn't have any tests for the optional fields, so added some

test/testsupport/testdata/TtpJsonResponses.scala
```diff
@@ -87,6 +88,20 @@ object TtpJsonResponses {
        |""".stripMargin
   }
 
+  def optionalEligibilityResponsesJson(eligibilityRules: EligibilityRules): String = {
```
AndyHWChung, line 91: you can perhaps save some repetition by having a method which looks something ilke this:
```scala
def optionalEligibilityResponsesJson(flag: Option[Boolean], flagName: String): String =
  flag.fold(""){f => s""",  "$flagName": ${f.toString} """   }
 ```
and then call it like:
```scala
optionalEligibilityResponsesJson(eligibilityRules.hasInvalidInterestSignals, "hasInvalidInterestSignals")
```
etc.

barnesjake: Sure thing, will update

AndyHWChung (dismissed):

AndyHWChung (approved):

PR #335, opened 2023-05-11
[JB][SUPP-767] add log messages to help with kibana dashboard

barnesjake:


app/services/TtpService.scala
```diff
@@ -86,6 +86,8 @@ class TtpService @Inject() (
         )
     }
     JourneyLogger.debug("EligibilityRequest: " + Json.prettyPrint(Json.toJson(eligibilityRequest)))
+    // below log message used by Kibana dashboard.
+    JourneyLogger.info(s"TTP eligibility check being made for ${journey.taxRegime.toString}")
```
AndyHWChung, line 90: do we want to maybe log the tax regime in the error message just below to be able to differentiate errors between different tax regimes?

barnesjake: We actually get that for free from the makeRishMessage in journey logger https://github.com/hmrc/essttp-frontend/blob/main/app/util/JourneyLogger.scala#L71 so don't think we need it

AndyHWChung: ah ok - cool 👍

AndyHWChung (approved):

PR #348, opened 2023-09-04
[JB][OPS-10902,OPS-10903] create new ineligible page for debt too low

barnesjake:


app/messages/Messages.scala
```diff
@@ -285,6 +285,41 @@ object Messages {
...
+      case TaxRegime.Vat =>
+        Message(
+          english = s"""You need to pay your VAT bill in full. Go to <a class="govuk-link" href="$link">GOV.UK</a> to make a payment today.""",
+          welsh   = s"""Mae angen i chi dalu’ch bil TAW yn llawn. Ewch i <a class="govuk-link" href="$link">GOV.UK</a> i wneud taliad heddiw."""
+        )
+    }
+
+    def `Call us on 0300 123 1813 if you are having difficulty making a payment online.`: Message = Message(
+      english = "Call us on <strong>0300 123 1813</strong> if you are having difficulty making a payment online.",
+      welsh   = "Os ydych yn cael anawsterau wrth dalu ar-lein, ffoniwch ni ar <strong>0300 200 1900</strong>."
```
barnesjake, line 320: call centre number is different for welsh, that's why the message differs slightly with the telephone number

PR #349, opened 2023-09-05
[JB][OPS-10902] tweak EligibilityRouter, I missed something from the …

barnesjake:
…ticket

### DIFF
--- a/app/controllers/pagerouters/EligibilityRouter.scala
+++ b/app/controllers/pagerouters/EligibilityRouter.scala
@@ -19,8 +19,8 @@ package controllers.pagerouters
 import controllers.routes
 import essttp.rootmodel.TaxRegime
 import essttp.rootmodel.ttp.eligibility.EligibilityCheckResult
-import models.EligibilityErrors
 import models.EligibilityErrors._
+import models.{EligibilityError, EligibilityErrors}
 import play.api.mvc.Call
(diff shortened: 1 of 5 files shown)


app/controllers/pagerouters/EligibilityRouter.scala
```diff
@@ -54,34 +55,43 @@ object EligibilityRouter {
...
     case TaxRegime.Vat   => routes.IneligibleController.vatDebtTooOldPage
   }
 
-  def whichFileYourReturnsPage(taxRegime: TaxRegime): Call = taxRegime match {
+  private def whichFileYourReturnsPage(taxRegime: TaxRegime): Call = taxRegime match {
     case TaxRegime.Epaye => routes.IneligibleController.epayeFileYourReturnPage
     case TaxRegime.Vat   => routes.IneligibleController.vatFileYourReturnPage
   }
 
+  //requirement from business that if multiple reasons exist but any of them are isLessThanMinDebtAllowance, go to debt too small page (PAYE only).
```
barnesjake, line 88: I realise this method probably seems a bit weird, but I wanted to keep the match statement above with the other checks on eligibility status a bit cleaner by not including the logic to check tax type so that's why this was added

AndyHWChung (approved):

PR #350, opened 2023-09-12
[JB][OPS-10926,OPS-10927] amend content on landing page and other rel…

barnesjake:
…evant pages for policy changes


app/messages/Messages.scala
```diff
@@ -468,9 +466,14 @@ object Messages {
       welsh   = s"mae arnoch ${maxAmountOfDebt.gdsFormatInPounds} neu lai"
     )
 
-    val `you do not have any other payment plans or debts with HMRC`: Message = Message(
-      english = "you do not have any other payment plans or debts with HMRC",
-      welsh   = "nid oes gennych unrhyw gynlluniau talu na dyledion eraill gyda CThEF"
+    val `your debts are 5 years old or less`: Message = Message(
+      english = "your debts are 5 years old or less",
```
AndyHWChung, line 470: it's a bit annoying that it's difficult to make it so that this content automatically updates when you update the value in config, but not sure there's a clean way around it 🤔

AndyHWChung: although you can passed in the config parameter to the mesages above and assumed it a value in years so maybe you could do the same here? But then the config value is to be taken as a value in years

DarrenCB: Is it possible to pass in a string from config? Like "5 years" or "28 days"?

AndyHWChung: you wouldn't be able to read it in as a `FiniteDuration` cos the longest unit they do there is days - you may be able to read it in with the unit separately as a Java time unit or something, but it may be a bit of faff to make it work

barnesjake: I did toy around with creating some adt's and using enumeratum's withNameFromString or whatever it was. It did work, but I thought it was a bit clunky as it would not fail at compilation if config was invalid so I reverted it to just using strings hardcoded in messages file

barnesjake: So for example I had config string under policy params for each regime that was
```
max-debt-age-type = "Monthly"
max-debt-age-value = 12
```

Which was read in AppConfig class and then the type derived was used in the Messages defs.



I did have a look at pureconfig library but thought it was overkill for this thing that won't be changing after it's released (in theory)

AndyHWChung: the most unambiguous thing to do which is also low effort would probably be to make the config value be in years so that what you've done in the other messages is true and then you could apply the same thing to here. This is only future-proof if any changes in the future has the max debt age in years still - if it doesn't then you'd need to change the config again and the messages. But at least then if you look at the config it's obvious that you definitely need to do more than a trivial change.

barnesjake: ok I'll update the config to reflect that it is for years

barnesjake: @AndyHWChung I've updated as per discussion. I think it could be good to have messages vary based on a string value in config, but as discussed, it's potentially more hassle than it's worth and a simple content update to the Messages file is all that's required if things change in future.

AndyHWChung: 👍 with the parameter from config in years now it'd be worth passing it in as a parameter to this message I think and replacing instances of `ageOfDebtNumber` with `ageOfDebtYears` or something just to make it really obvious

barnesjake: ah yeah good point, I'll update, annoyingly VAT won't be changing to use years for a little while (hopefully soon after talking to Karen) so I won't be able to do a blanket swap, but will update where I can

barnesjake: ah right sorry I'm following now, updated

AndyHWChung (approved):

PR #351, opened 2023-09-13
[JB][OPS-10984] send vat users to debt too small when they have multi…

barnesjake:
…ple reasons and one is isLessThanMinDebtAllowance


barnesjake:
This one is annoying, I initially wanted to remove `private def determineWhetherToGoToDebtTooSmall`, but having the guard in the pattern match made the formatting grim (i.e. all the `=>` wouldn't be aligned), so I opted for keeping it

AndyHWChung (approved):

PR #352, opened 2023-09-13
[JB][OPS-10999] derive max plan length from config instead of hard coded

barnesjake:


AndyHWChung:
> * `uk.gov.hmrc.sbt-auto-build` can be updated to `3.13.0` in [project/plugins.sbt](https://github.com/hmrc/essttp-frontend/blob/jake%2FOPS-10999/project/plugins.sbt#L13)

this would be a quick thing to sort out while you're here

AndyHWChung (approved):

PR #368, opened 2023-11-22
[OPS-11321] Add kick out page for VAT chargesBeforeMaxAccountingDate eligibility rule

Kntanos:


app/config/AppConfig.scala
```diff
@@ -117,6 +117,7 @@ class AppConfig @Inject() (config: Configuration, servicesConfig: ServicesConfig
       val maxPlanDurationInMonths: Int = getParam[Int]("max-plan-duration-in-months")
       val maxAgeOfDebtInDays: Int = getParam[Int]("max-age-of-debt-in-days")
       val payOnlineLink: String = getParam[String]("pay-online-link")
+      val vatAccountingPeriodStart: String = getParam[String]("vat-accounting-period-start")
```
AndyHWChung, line 120: it'd be nicer if this was a `LocalDate` at this point so we're passing around more well-typed values further on I think. Something like this maybe?
```scala
val vatAccountingPeriodStart: LocalDate = {
  val string = getParam[String]("vat-accounting-period-start")
  LocalDate.parse(string) // may need a formatter here?
}
```

Kntanos: Done. Thanks

AndyHWChung (approved):

PR #384, opened 2024-01-26
[KN][OPS-11464] CR111 - Updated Your Bill page

Kntanos:


app/controllers/YourBillController.scala
```diff
@@ -85,6 +86,14 @@ object YourBillController {
     InvoicePeriod(monthNumberInTaxYear(startDate), startDate, endDate, dueDate)
   }
 
+  private def chargeBearsInterest(ass: ChargeTypeAssessment): Option[IsInterestBearingCharge] = {
+    List(ass).headOption.map { (chargeTypeAssessment: ChargeTypeAssessment) =>
```
AndyHWChung, line 90: `List(ass).headOption` will just return `ass` - couldn't you just do this?
```scala
ass.charges.exists(_.isInterestBearingCharge)
```

Kntanos: Not this, but yes. Corrected. Thank you

PR #406, opened 2024-03-15
[KN][OPS-11901] Update file your return kick out page

Kntanos:


app/messages/Messages.scala
```diff
@@ -500,24 +500,14 @@ object Messages {
...
-
-    val `If you have recently filed your return, your account may take up to 72 hours to be updated before you can set up a payment plan.`: Message = Message(
-      english = "If you have recently filed your return, your account may take up to 72 hours to be updated before you can set up a payment plan.",
-      welsh   = "Os ydych chi wedi cyflwyno’ch Ffurflen Dreth yn ddiweddar, gallai gymryd hyd at 72 awr i’ch cyfrif gael ei ddiweddaru cyn i chi allu trefnu cynllun talu."
-    )
-
-    def `You must file your tax return before you can set up a Self Assessment payment plan online.`(fileSaReturnUrl: String): Message = Message(
-      english = s"""You must <a class="govuk-link" href="$fileSaReturnUrl">file your tax return</a> before you can set up a Self Assessment payment plan online.""",
-      welsh   = s"""Mae’n rhaid i chi <a class="govuk-link" href="$fileSaReturnUrl">gyflwyno’ch Ffurflen Dreth</a> cyn i chi allu trefnu cynllun talu ar gyfer Hunanasesiad ar-lein."""
+    def `You must file your tax return before you can set up a ... payment plan online`(fileReturnUrl: String, taxRegimeFullName: (String, String)): Message = Message(
```
AndyHWChung, line 508: it'd be nicer to pass in `TaxRegime` to this method instead of `(String, String)` I think - it'd be easier to track what the full messages are cos it'd all be in one place rather than the content being split outside of this file

Kntanos: You mean having three cases for the content of each regime, like it was before?

AndyHWChung (approved):

PR #413, opened 2024-03-22
Kostas/ops 11678

Kntanos:


conf/application.conf
```diff
@@ -170,7 +170,6 @@ contact-frontend {
 
 features {
   email-journey = true
-  vat = true
   sa = true
   cr111 = true
```
AndyHWChung, line 174: can't we get rid of this one now as well?

Kntanos: Duh... missed it 👍🏻

AndyHWChung (approved):

PR #417, opened 2024-04-02
[KN][OPS-11957] Add new kick out after ddInProgress

Kntanos:


Kntanos:
test me

app/messages/Messages.scala
```diff
@@ -600,7 +610,7 @@ object Messages {
     )
 
     def `I do not want to set up a payment plan`(link: String): Message = Message(
-      english = s"""<a class="govuk-link" href="$link">I do not want to set up a payment plan</a>""",
+      english = s"""<a class="govuk-link" id="kickout" href="$link">I do not want to set up a payment plan</a>""",
```
AndyHWChung, line 613: if we have the id in english it might make sense to have it for the welsh as well for consistency in case we ever want to retrieve the same element by the same id in welsh

Kntanos: of course

AndyHWChung: thanks!

AndyHWChung (approved):

PR #421, opened 2024-04-10
OPS-11996 - MIS for Dual charging - update/add CIP events

DreamcastKnight:


app/services/AuditService.scala
```diff
@@ -207,6 +214,23 @@ class AuditService @Inject() (auditConnector: AuditConnector)(implicit ec: Execu
     )
   }
 
+  private def toDdinProgressAuditDetail(journey: Journey, continueOrExitBool: Boolean)
+    (implicit r: AuthenticatedJourneyRequest[_]): DdInProgressAuditDetail = {
+    val continueOrExit: String = continueOrExitBool match {
```
AndyHWChung, line 219: it'd probably be a cleaner to just to `if(continueOrExitBool) "continue" else "exit"` since it's just a boolean. Also, it's nice when the name of the boolean tells you what true or false means to avoid ambiguity - maybe something like `hasChosenToContinue` would convey that?

app/controllers/IneligibleController.scala
```diff
@@ -185,10 +187,12 @@ class IneligibleController @Inject() (
     )
 
   val epayeYouHaveChosenNotToSetUpPage: Action[AnyContent] = as.authenticatedJourneyAction { implicit request =>
+    auditService.auditDdInProgress(request.journey, false)
```
AndyHWChung, line 190: it'd be nice to add in something in the unit tests which checks that the correct thing is audited whenever you trigger an event like this. Here, in the spec which checks the content of the page you could add in something there for instance. There should be something in the test `AuditStub` already which already allows you to check that an audit event looks correct.

test/controllers/IneligibleControllerSpec.scala
```diff
@@ -359,6 +360,25 @@ class IneligibleControllerSpec extends ItSpec {
...
+        "DirectDebitInProgress",
+        Json.parse(
+          s"""
+             |{
+             |  "origin": "Bta",
+             |  "taxType": "Vat",
+             |  "taxDetail": $expectedTaxDetailsJson,
+             |  "correlationId": "CorrelationId(8d89a98b-0b26-4ab2-8114-f7c7c81c3059)",
+             |  "authProviderId": "GGCredId(authId-999)",
+             |  "continueOrExit": "exit"
```
AndyHWChung, line 376: there's the case where we get `"continue"` where it'd be nice to have a unit test as well in the appropriate place

AndyHWChung (approved):

PR #424, opened 2024-05-13
Kostas/ops 11749

Kntanos:


test/controllers/YourBillControllerSpec.scala
```diff
@@ -157,20 +158,22 @@ class YourBillControllerSpec extends ItSpec {
       tableRows(1).select(".govuk-summary-list__value").text() shouldBe "£1,000"
     }
 
-    "return your bill page for SA" in {
+    "return your bill page for SA for known MainTrans code" in {
```
AndyHWChung, line 161: it'd be nice to have a test for each main trans code if we can just so we've got the coverage on the conversion from code to content on the page somewhere

README.md
```diff
@@ -79,6 +79,57 @@ see here for more BARs stub data https://github.com/hmrc/bank-account-reputation
 
 ---
 
+### MainTrans codes for SA
+In SA we look up the MainTrans code received in ChargeReference in the table below and display the corresponding charge
+type in the 'You Bill' page. If the code does not have a corresponding charge description, the user is take to the 
+generic sa ineligibility kick out page.
+
+| Main Trans code | English description                            | Welsh description                                   |
```
AndyHWChung, line 87: it might be better to have this table sitting in confluence somewhere and linking to that. If you were looking for the rules confluence is probably where the wider audience would look maybe? Maybe see what Karen thinks to she if she thinks there's a good place for the table there

app/controllers/YourBillController.scala
```diff
@@ -51,13 +51,21 @@ class YourBillController @Inject() (
...
+      Ok(
+        views.yourBillIs(
+          YourBillController.overDuePayments(journey.eligibilityCheckResult),
+          journey.taxRegime
+        )
       )
-    )
+    } catch {
+      // SA only: if MainTrans is not found, see README
+      case e: NoSuchElementException =>
```
AndyHWChung, line 64: it might be safer to use and easier to trace if we had a specific exception for this just in case something is added in the future which can also give a `NoSuchElementException`, we could have something like this for example:
```scala
object MainTrans {
  case object UnknownMainTrans extends Throwable
}

....
// where we're getting from the ChargeTypeMessages map
val message = ChargeTypeMessages.chargeFromMTrans.getOrElse(
  mTrans,
  throw new UnknownMainTrans
)
```

AndyHWChung (approved):

PR #427, opened 2024-05-17
Kostas/ops 12020

Kntanos:


app/connectors/CallEligibilityApiRequest.scala
```diff
@@ -16,17 +16,51 @@
...
+import essttp.rootmodel.ttp.eligibility.{IdType, IdValue, Identification}
+import models.EligibilityReqIdentificationFlag
+import play.api.libs.json._
+import play.api.libs.functional.syntax._
 
 final case class CallEligibilityApiRequest(
     channelIdentifier:         String,
-    idType:                    String,
-    idValue:                   String,
+    identification:            List[Identification],
```
AndyHWChung, line 26: does the stub need updating as well to be able to handle both forms of the request?

app/connectors/CallEligibilityApiRequest.scala
```diff
@@ -16,17 +16,51 @@
...
 
 object CallEligibilityApiRequest {
   @SuppressWarnings(Array("org.wartremover.warts.Any"))
-  implicit val format: OFormat[CallEligibilityApiRequest] = Json.format[CallEligibilityApiRequest]
+  implicit def format(implicit e: EligibilityReqIdentificationFlag): Format[CallEligibilityApiRequest] = {
+    val writes = if (e.value) {
+      Json.writes[CallEligibilityApiRequest]
+    } else {
+      OWrites { callEligibilityApiRequest: CallEligibilityApiRequest =>
+        val firstIdentification = callEligibilityApiRequest.identification.headOption
```
AndyHWChung, line 38: could we maybe make it explode if we're trying to condense a list with more than one thing? If we somehow weer in that situation we'd probably want to know cos it wouldn't be right unless we were looking for a specific identifier which we can't code for yet. You could pattern match for this if you want:
```scala
list match {
  case Nil => nothing there
  case head :: Nil => // only one thing
  case other => // more than one thing
}
```

Kntanos: Throw exceptions for Nil and other?

AndyHWChung: that's what I was thinking yeah - with each having it's own specific message

AndyHWChung (approved):

PR #432, opened 2024-06-11
[JG] OPS-12246 - confirmation and print payment page changes

DreamcastKnight:


app/views/SaPrintSummaryPage.scala.html
```diff
@@ -0,0 +1,286 @@
...
+@layout(pageTitle = Some(title), additionalScripts = Some(scripts)) {
+
+    <h1 class="govuk-heading-xl">
+    @title
+    </h1>
+
+    <p class="govuk-body js-visible">
+        <a href="#print-dialogue" class="govuk-link" id="printLink">
+            @{
+            Messages.Confirmation.`Print a copy of your payment plan`.show
```
Kntanos, line 142: I think this is not consistent with the other two regimes. You may want to raise it and make it consistent or correct this to match the prototype. It is meant to say `Print or save a copy of your payment plan`

Kntanos (approved):

PR #448, opened 2024-08-29
[KN][OPS-12588] Handle PEGA going back to 'can you pay within 6 month…

Kntanos:
…s' page

### DIFF
--- a/app/controllers/CanPayWithinSixMonthsController.scala
+++ b/app/controllers/CanPayWithinSixMonthsController.scala
@@ -68,10 +68,6 @@ class CanPayWithinSixMonthsController @Inject() (
       }
     }
 
-  def backFromPegaLanding(regime: TaxRegime): Action[AnyContent] = as.authenticatedJourneyAction { _ =>
-    Redirect(controllers.routes.CanPayWithinSixMonthsController.canPayWithinSixMonths(regime)).withNewSession
-  }
-
   val canPayWithinSixMonthsSubmit: Action[AnyContent] = as.authenticatedJourneyAction.async { implicit request =>
     CanPayWithinSixMonthsForm.form.bindFromRequest().fold(
       formWithErrors => Ok(views.canPayWithinSixMonthsPage(formWithErrors, remainingAmountToPay(request.journey))),
--- a/app/controllers/CanPayWithinSixMonthsController.scala
+++ b/app/controllers/CanPayWithinSixMonthsController.scala
@@ -22,7 +22,7 @@ import controllers.JourneyFinalStateCheck.finalStateCheck
 import essttp.journey.JourneyConnector
 import essttp.journey.model.CanPayWithinSixMonthsAnswers.CanPayWithinSixMonths
 import essttp.journey.model.{CanPayWithinSixMonthsAnswers, Journey, UpfrontPaymentAnswers}
-import essttp.rootmodel.{AmountInPence, UpfrontPaymentAmount}
+import essttp.rootmodel.{AmountInPence, TaxRegime, UpfrontPaymentAmount}
 import models.enumsforforms.CanPayWithinSixMonthsFormValue
 import models.forms.CanPayWithinSixMonthsForm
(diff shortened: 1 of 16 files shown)


app/controllers/CanPayWithinSixMonthsController.scala
```diff
@@ -47,23 +48,28 @@ class CanPayWithinSixMonthsController @Inject() (
...
+              val form = previousAnswers.fold(CanPayWithinSixMonthsForm.form)(value =>
+                CanPayWithinSixMonthsForm.form.fill(CanPayWithinSixMonthsFormValue.canPayWithinSixMonthsToFormValue(value)))
+
+              Ok(views.canPayWithinSixMonthsPage(form, remainingAmountToPay(j)))
+            }
+          )
+      }
     }
+
+  def backFromPegaLanding(regime: TaxRegime): Action[AnyContent] = as.authenticatedJourneyAction { _ =>
```
AndyHWChung, line 71: you shouldn't need this endpoint I think - they would just point directly to the `canPayWithinSixMonths` endpoint

AndyHWChung: oh wait, I think I see what this is now. This should go in the test-only pega controller since we don't want to expose this in the actual service

test/controllers/CanPayWithinSixMonthsControllerSpec.scala
```diff
@@ -73,7 +74,7 @@ class CanPayWithinSixMonthsControllerSpec extends ItSpec {
       stubCommonActions()
       EssttpBackend.Dates.findJourneyExtremeDates(testCrypto, Origins.Epaye.Bta)()
 
-      val result = controller.canPayWithinSixMonths(fakeRequest)
+      val result = controller.canPayWithinSixMonths(Epaye)(fakeRequest)
```
AndyHWChung, line 77: since this endpoint is using the common reconstruct session stuff for PEGA you could re-use some of the specs I wrote like this:
https://github.com/hmrc/essttp-frontend/blob/main/test/controllers/PaymentScheduleControllerSpec.scala#L328

Those are error tests around the behaviour of reconstructing.  It'd be nice to get a test where we need to reconstruct the session and it's successful - kind of like this:

https://github.com/hmrc/essttp-frontend/blob/main/test/controllers/PaymentScheduleControllerSpec.scala#L433

conf/testOnlyDoNotUseInAppConf.routes
```diff
@@ -13,53 +13,54 @@
...
+GET  /set-up-a-payment-plan/test-only/start-journey-sa-bta              testOnly.controllers.StartJourneyController.startJourneySaBta
+GET  /set-up-a-payment-plan/test-only/start-journey-sa-pta              testOnly.controllers.StartJourneyController.startJourneySaPta
+GET  /set-up-a-payment-plan/test-only/start-journey-sa-mobile           testOnly.controllers.StartJourneyController.startJourneySaMobile
 
 
-GET  /set-up-a-payment-plan/test-only/pega                         testOnly.controllers.PegaController.dummyPegaPage(regime: TaxRegime)
-POST /set-up-a-payment-plan/test-only/pega                         testOnly.controllers.PegaController.dummyPegaPageContinue(regime: TaxRegime)
+GET  /set-up-a-payment-plan/test-only/pega                              testOnly.controllers.PegaController.dummyPegaPage(regime: TaxRegime)
+POST /set-up-a-payment-plan/test-only/pega                              testOnly.controllers.PegaController.dummyPegaPageContinue(regime: TaxRegime)
+GET  /set-up-a-payment-plan/test-only/paying-within-six-months/back     testOnly.controllers.PegaController.backFromPegaLanding(regime: TaxRegime)
```
AndyHWChung, line 64: this would be more consistent with the other pega endpoints we have:
```suggestion
GET  /set-up-a-payment-plan/test-only/pega/back     testOnly.controllers.PegaController.backFromPegaLanding(regime: TaxRegime)
```

AndyHWChung (approved):

PR #449, opened 2024-09-11
[JG] OPS-12345 - update firestarter page for simple assessment

DreamcastKnight:


app/config/AppConfig.scala
```diff
@@ -84,6 +85,10 @@ class AppConfig @Inject() (config: Configuration, servicesConfig: ServicesConfig
       val baseUrl = BaseUrl.platformHost.getOrElse("http://localhost:9063")
       s"$baseUrl/pay-what-you-owe-in-instalments"
     }
+    val siaSuppUrl: String = {
```
AndyHWChung, line 88: we don't need this - the other service doesn't support SIA

app/controllers/DetermineTaxIdController.scala
```diff
@@ -69,6 +69,7 @@ class DetermineTaxIdController @Inject() (
         case TaxRegime.Epaye => Redirect(routes.NotEnrolledController.notEnrolled)
         case TaxRegime.Vat   => Redirect(routes.NotEnrolledController.notVatRegistered)
         case TaxRegime.Sa    => Redirect(routes.NotEnrolledController.notSaEnrolled)
+        case TaxRegime.Sia   => Redirect(routes.NotEnrolledController.notSaEnrolled)
```
AndyHWChung, line 72: this doesn't seem quite right - if the user didn't have a NINO then I'm assuming that there'd be some other new special kickout page we could point to here which we could just have as a dummy endpoint for now

app/controllers/LandingController.scala
```diff
@@ -69,6 +69,20 @@ class LandingController @Inject() (
...
+    if (appConfig.siaEnabled) {
+      as.default.async { implicit request =>
+        checkNotShuttered(TaxRegime.Sia) {
+          getBackUrl().map { maybeBackUrl =>
+            Ok(views.siaLanding(maybeBackUrl))
+          }
+        }
+      }
+    } else {
+      as.default(_ => Redirect(appConfig.Urls.siaSuppUrl))
```
AndyHWChung, line 82: here we can just blow up if it's not enabled

app/controllers/DetermineTaxIdController.scala
AndyHWChung: in the pattern match above here:
```scala
 case j: Journey.Stages.Started =>
        enrolmentService.determineTaxIdAndUpdateJourney(j, request.enrolments) -> j.taxRegime
``` 
I think you're gonna have to do something different if it's SIA since SIA doesn't have an enrolment. What you'd need to do is to make sure we're retrieving the NINO from auth and make it available in the `request` here and then use that as the tax id if it exists.

app/controllers/WhichTaxRegimeController.scala
```diff
@@ -49,23 +49,25 @@ class WhichTaxRegimeController @Inject() (
       val hasSaEnrolment = EnrolmentDef.Sa.findEnrolmentValues(request.enrolments).isSuccess
 
       (hasEpayeEnrolment, hasVatEnrolment, hasSaEnrolment) match {
-        case (true, false, false)                        => Redirect(routes.LandingController.epayeLandingPage)
-        case (false, true, false)                        => Redirect(routes.LandingController.vatLandingPage)
-        case (false, false, true) if appConfig.saEnabled => Redirect(routes.LandingController.saLandingPage)
+        case (true, false, false)                          => Redirect(routes.LandingController.epayeLandingPage)
+        case (false, true, false)                          => Redirect(routes.LandingController.vatLandingPage)
+        case (false, false, true) if appConfig.saEnabled   => Redirect(routes.LandingController.saLandingPage)
+        case (false, false, false) if appConfig.siaEnabled => Redirect(routes.LandingController.siaLandingPage)
```
AndyHWChung, line 55: we're gonna fix this up in a later ticket so we don't need to worry about the which tax regime stuff for this ticket

app/messages/Messages.scala
```diff
@@ -156,6 +161,12 @@ object Messages {
             english = s"Your Self Assessment tax bill is ${amount.gdsFormatInPounds}",
             welsh   = s"Mae’ch bil treth Hunanasesiad yn dod i gyfanswm o ${amount.gdsFormatInPounds}"
           )
+
+        case TaxRegime.Sia =>
+          Message(
+            english = s"Your Simple Assessment tax bill is ${amount.gdsFormatInPounds}",
```
AndyHWChung, line 167: We should get this content checked - not necessarily int this ticket,  but we should note it down in that ticket we were gonna list things we were unsure of in. Same goes for the rest of the content you've added which isn't specifically mentioned on the ticket.

app/messages/Messages.scala
```diff
@@ -628,7 +690,8 @@ object Messages {
         english = "You already have a Direct Debit set up for VAT.",
         welsh   = "Mae eisoes gennych drefniant Debyd Uniongyrchol er mwyn talu TAW."
       )
-      case TaxRegime.Sa => throw new NotImplementedError("ddInProgress flag not relevant to SA charges")
+      case TaxRegime.Sa  => throw new NotImplementedError("ddInProgress flag not relevant to SA charges")
+      case TaxRegime.Sia => throw new NotImplementedError("ddInProgress flag not relevant to SIA charges")
```
AndyHWChung, line 694: we should double check this is correct for SIA

app/messages/Messages.scala
```diff
@@ -1657,6 +1726,11 @@ object Messages {
             english = "Debt Management<br>Self Assessment<br>HM Revenue and Customs<br>BX9 1AS<br>United Kingdom",
             welsh   = "Rheolaeth Dyledion<br>Hunanasesiad<br>Gwasanaeth Cwsmeriaid Cymraeg CThEF<br>HMRC<br>BX9 1ST"
           )
+        case TaxRegime.Sia =>
+          Message(
+            english = "Debt Management<br>Simple Assessment<br>HM Revenue and Customs<br>BX9 1AS<br>United Kingdom",
```
AndyHWChung, line 1731: this we should definitely double check

app/services/EnrolmentService.scala
```diff
@@ -58,6 +58,12 @@ class EnrolmentService @Inject() (journeyService: JourneyService, auditService:
           EnrolmentDef.Sa.findEnrolmentValues(enrolments),
           journey
         )(journeyService.UpdateTaxRef.updateSaTaxId(journey.id, _))
+
+      case TaxRegime.Sia =>
```
AndyHWChung, line 62: this could blow up if we want if we make sure we don't call this service for SIA

app/testOnly/models/formsmodel/StartJourneyForm.scala
```diff
@@ -159,6 +161,7 @@ object StartJourneyForm {
           case TaxRegime.Epaye => payeTaxReferenceKey -> RandomDataGenerator.nextEpayeRefs()(Random)._3
           case TaxRegime.Vat   => vatTaxReferenceKey -> RandomDataGenerator.nextVrn()(Random)
           case TaxRegime.Sa    => saTaxReferenceKey -> RandomDataGenerator.nextSaUtr()(Random)
+          case TaxRegime.Sia   => saTaxReferenceKey -> RandomDataGenerator.nextSaUtr()(Random)
```
AndyHWChung, line 164: this should be referring to NINO's instead of SA UTR. The stub will probably also need to be updated as well.
I think you'll need to add NINO in the form object and then you'd need to go into `TestUser`  and make sure we're creating a user there with the supplied NINO

app/testOnly/views/IAmGovUkPage.scala.html
```diff
@@ -36,6 +36,7 @@
         case TaxRegime.Epaye => routes.StartJourneyController.startGovukEpayeJourney
         case TaxRegime.Vat => routes.StartJourneyController.startGovukVatJourney
         case TaxRegime.Sa => routes.StartJourneyController.startGovukSaJourney
+        case TaxRegime.Sia => routes.StartJourneyController.startGovukSaJourney
```
AndyHWChung, line 39: looks like this should be `startGovukSiaJourney`

app/views/YouAlreadyHaveDirectDebit.scala.html
```diff
@@ -45,6 +45,7 @@
         case TaxRegime.Epaye => routes.IneligibleController.epayeYouHaveChosenNotToSetUpPage.url
         case TaxRegime.Vat => routes.IneligibleController.vatYouHaveChosenNotToSetUpPage.url
         case TaxRegime.Sa => throw new NotImplementedError("SA does not receive ddInProgress flag")
+        case TaxRegime.Sia => throw new NotImplementedError("SIA does not receive ddInProgress flag")
```
AndyHWChung, line 48: we should get this double checked

app/views/partials/NoDueDatesReachedPartial.scala.html
```diff
@@ -33,6 +33,7 @@
     case TaxRegime.Epaye => Messages.NotEligible.`You cannot set up an Employers' PAYE payment plan online...not overdue`
     case TaxRegime.Vat => Messages.NotEligible.`You cannot set up a VAT payment plan online...not overdue`
     case TaxRegime.Sa => throw new NotImplementedError()
+    case TaxRegime.Sia => throw new NotImplementedError()
```
AndyHWChung, line 36: should get this double checked as well

app/views/partials/OverduePaymentRow.scala.html
```diff
@@ -42,6 +42,9 @@
                     s"${Messages.YourBill.to.show} ${invoicePeriodDateShortMonth(payment.invoicePeriod.end)} ${payment.invoicePeriod.end.getYear.toString}"
         case TaxRegime.Sa =>
             Messages.YourBill.`Due`(payment.invoicePeriod.dueDate).show
+        case TaxRegime.Sia =>
+            s"${invoicePeriodDateShortMonth(payment.invoicePeriod.start)} " + s"${invoicePeriodDateYear(payment.invoicePeriod.start)} " +
```
AndyHWChung, line 46: this will eventually need double checking since the format can be different per tax regime

test/controllers/EmailControllerSpec.scala
```diff
@@ -452,7 +456,7 @@ class EmailControllerSpec extends ItSpec {
                 Json.parse(
                   s"""
                    |{
-                   |  "origin" : "Bta",
+                   |  "origin" : "${origin.toString().takeRight(3)}",
```
AndyHWChung, line 459: this might not be appropriate for other taxr egimes in the future - if you look at how the audit event is populating this value then you'd be able to do the same in the test here. This shoul be changed where it appears in other specs as well

test/controllers/IneligibleControllerSpec.scala
```diff
@@ -62,7 +64,9 @@ class IneligibleControllerSpec extends ItSpec {
             case TaxRegime.Epaye => Some(Set(TdAll.payeEnrolment))
             case TaxRegime.Vat   => Some(Set(TdAll.vatEnrolment))
             case TaxRegime.Sa    => Some(Set(TdAll.saEnrolment))
+            case TaxRegime.Sia   => Some(Set(TdAll.saEnrolment))
```
AndyHWChung, line 67: here you'd just set enrolments to `None` here or  `Some(Set.empty)` depending on how it's being used later on

test/controllers/PaymentPlanSetUpControllerSpec.scala
```diff
@@ -42,7 +42,7 @@ class PaymentPlanSetUpControllerSpec extends ItSpec {
 
   List(
     Origins.Epaye.Bta,
-    Origins.Vat.Bta,
+    Origins.Vat.Bta
```
AndyHWChung, line 45: does this need SA and SIA adding here?

DreamcastKnight: sa has it's own section, and currently sia doesn't really have any data to check

AndyHWChung: if the tests for SA are doing the same thing as the ones for Epaye and Vat here then it might be worth exploring just adding in the tax regime to this list. What happens if you add those regimes to the list? Does it break?

DreamcastKnight: it breaks; I believe because sa has something different to the other two

AndyHWChung: ah ok - is it possible to write SIA tests or do we need something else in place first?

app/actions/AuthenticatedActionRefiner.scala
```diff
@@ -46,16 +46,18 @@ trait AuthenticatedActionRefiner { this: ActionRefiner[Request, AuthenticatedReq
...
+        case enrolments ~ credentials ~ nino =>
           credentials match {
             case None =>
               Future.failed(new RuntimeException(s"Could not find credentials"))
 
             case Some(ggCredId) =>
               Future.successful(
-                Right(new AuthenticatedRequest[A](request, enrolments, GGCredId(ggCredId.providerId)))
+                Right(
+                  new AuthenticatedRequest[A](request, enrolments, GGCredId(ggCredId.providerId), Some(Nino(nino.getOrElse(""))))
```
AndyHWChung, line 59: it'd make more sense to map into `nino` here so that we get a `None` if `nino` is also `None`

app/controllers/DetermineTaxIdController.scala
```diff
@@ -55,7 +55,11 @@ class DetermineTaxIdController @Inject() (
   def determineTaxId(): Action[AnyContent] = as.authenticatedJourneyAction.async { implicit request =>
     val (maybeTaxId, taxRegime): (Future[Option[TaxId]], TaxRegime) = request.journey match {
       case j: Journey.Stages.Started =>
-        enrolmentService.determineTaxIdAndUpdateJourney(j, request.enrolments) -> j.taxRegime
+        if (j.taxRegime == TaxRegime.Sia) {
+          (Future(request.nino.asInstanceOf[Option[TaxId]]), j.taxRegime)
```
AndyHWChung, line 59: isn't `Nino` already a `TaxId`? Do you need this `.asInstanceOf`?

app/controllers/DetermineTaxIdController.scala
```diff
@@ -69,6 +73,7 @@ class DetermineTaxIdController @Inject() (
         case TaxRegime.Epaye => Redirect(routes.NotEnrolledController.notEnrolled)
         case TaxRegime.Vat   => Redirect(routes.NotEnrolledController.notVatRegistered)
         case TaxRegime.Sa    => Redirect(routes.NotEnrolledController.notSaEnrolled)
+        case TaxRegime.Sia   => Redirect(routes.NotEnrolledController.notSaEnrolled)
```
AndyHWChung, line 76: perhaps we should just throw an exception here for now instead of going to a page for a different tax regime. I'm not sure if we're gonna have a dedicated page for this scenario actually, so it'd be the least amount of work in the future to throw here if we end up not having one

AndyHWChung: oh wait a sec, looks like you've got a `siaNoNino` endpoint already - I guess this should point to that endpoint instead. It looks like we are having a kickout page having looked at another ticket

app/testOnly/controllers/StartJourneyController.scala
```diff
@@ -188,6 +211,11 @@ class StartJourneyController @Inject() (
     withSessionId(Future.successful(Ok(iAmGovUkPage(TaxRegime.Sa))))
   }
 
+  /** Pretends being a Govuk Sa page */
```
AndyHWChung, line 214: `Sia` not `Sa` here in the comments

test/controllers/PaymentPlanSetUpControllerSpec.scala
DreamcastKnight: for now there wasnt even an siapaymentplan page; I have realised there probably should be something there, so I have made one real quick. for now it is basically identical to the sa page (until we know what it should be like).  adding a test in just so it's not forgotten when we get to changing the page

app/testOnly/views/TestOnlyStartPage.scala.html
```diff
@@ -724,6 +724,117 @@
...
+            content = Text(e.message)
+        )),
+        value = form.data.get("siaDebtTotalAmount").orElse(Some("10000")),
+        hint = Some(Hint(content = Text(s"Enter an amount between £1 and ${appConfig.PolicyParameters.SIA.maxAmountOfDebt.gdsFormatInPounds} for SIA.")))
+    ))
+
+    @interestAmountContent
+
+    @govukInput(Input(
+        id = "siaNINO",
```
AndyHWChung, line 756: I'm not sure the `id` and `name` here are consistent with what you've written in the actual form logic. In the form you're looking for `siaTaxReference` which is equal to `siaTaxReference`. Perhaps it could be more generic with some like `nino` since we could in theory be looking for nino's for other tax regimes in the future

conf/application.conf
```diff
@@ -172,6 +172,7 @@ contact-frontend {
 features {
   email-journey = true
   sa = true
+  sia = true
```
AndyHWChung, line 175: we should raise an app-config-prod PR as well to disable this flag to make sure we don't accidentally deploy with SIA enabled there

test/controllers/IneligibleControllerSpec.scala
```diff
@@ -62,7 +64,9 @@ class IneligibleControllerSpec extends ItSpec {
             case TaxRegime.Epaye => Some(Set(TdAll.payeEnrolment))
             case TaxRegime.Vat   => Some(Set(TdAll.vatEnrolment))
             case TaxRegime.Sa    => Some(Set(TdAll.saEnrolment))
+            case TaxRegime.Sia   => None
           }
+          //OPS-12345 - sia doesn't have an enrolment; uncertain what should go here, but the match needs an option for sia
```
AndyHWChung, line 69: does this comment still apply?

test/controllers/IneligibleControllerSpec.scala
```diff
@@ -486,7 +509,9 @@ class IneligibleControllerSpec extends ItSpec {
             case TaxRegime.Epaye => Some(Set(TdAll.payeEnrolment))
             case TaxRegime.Vat   => Some(Set(TdAll.vatEnrolment))
             case TaxRegime.Sa    => Some(Set(TdAll.saEnrolment))
+            case TaxRegime.Sia   => Some(Set(TdAll.saEnrolment))
           }
+          // sia has no enrolment, but this match needed an option for it
```
AndyHWChung, line 514: you could have `None` for `Sia` here like before

test/controllers/PaymentScheduleControllerSpec.scala
```diff
@@ -264,7 +265,7 @@ class PaymentScheduleControllerSpec extends ItSpec with PegaRecreateSessionAsser
                   s"""
                  |{
                  |        "correlationId": "8d89a98b-0b26-4ab2-8114-f7c7c81c3059",
-                 |        "origin": "Bta",
+                 |        "origin": "${origin.toString().takeRight(3)}",
```
AndyHWChung, line 268: this logic could be adapted to look more like the `AuditService` logic like you did in another test

app/controllers/DetermineTaxIdController.scala
DreamcastKnight: originally it was complaining for not doinf the instance of, but you're right, doesnt need to be there

test/controllers/SubmitArrangementControllerSpec.scala
```diff
@@ -145,7 +148,7 @@ class SubmitArrangementControllerSpec extends ItSpec {
                      |	},
                      |	"status": "successfully sent to TTP",
                      |	"failedSubmissionReason": 202,
-                     |	"origin": "Bta",
+                     |	"origin": "${origin.toString().takeRight(3)}",
```
AndyHWChung, line 151: this could be fixed up as well

app/controllers/DetermineTaxIdController.scala
DreamcastKnight: yeah, I just forgot to change this to point to the no nino endpoint, mb

app/testOnly/models/formsmodel/StartJourneyForm.scala
```diff
@@ -89,10 +89,12 @@ object StartJourneyForm {
   private val payeDebtTotalAmountKey: String = "payeDebtTotalAmount"
   private val vatDebtTotalAmountKey: String = "vatDebtTotalAmount"
   private val saDebtTotalAmountKey: String = "saDebtTotalAmount"
+  private val siaDebtTotalAmountKey: String = "siaDebtTotalAmount"
 
   private val payeTaxReferenceKey: String = "payeTaxReference"
   private val vatTaxReferenceKey: String = "vatTaxReference"
   private val saTaxReferenceKey: String = "saTaxReference"
+  private val siaTaxReferenceKey: String = "siaTaxReference"
```
AndyHWChung, line 97: wouldn't this be `nino` now that you've changed the id in the template? You'd be able to tell if it's working by entering a NINO on the start page and checking the journey in mongo has the NINO you typed in once it's created

AndyHWChung (approved):

PR #453, opened 2024-09-19
[JG] OPS-12345 - addressing comments on ticket

DreamcastKnight:


app/testOnly/views/TestOnlyStartPage.scala.html
```diff
@@ -745,9 +745,7 @@
         )),
         errorMessage = form.error("siaDebtTotalAmount").map(e => ErrorMessage.errorMessageWithDefaultStringsTranslated(
             content = Text(e.message)
-        )),
```
AndyHWChung, line 748: isn't what you've removed here still relevant? Doesn't it get used to stick into the eligibility result in the stubs?

DreamcastKnight: it could be, it was one of the bits that jamie pointed out, but if it should be there I can add it back in

app/controllers/pagerouters/EligibilityRouter.scala
```diff
@@ -66,8 +66,10 @@ object EligibilityRouter {
         case Some(DmSpecialOfficeProcessingRequiredCDCS) =>
           if (taxRegime === Sa) whichGenericIneligiblePage(taxRegime)
           else throw new NotImplementedError(s"dmSpecialOfficeProcessingRequiredCDCS ineligibility reason not relevant to ${taxRegime.entryName}")
-        case Some(IsAnMtdCustomer)                       => whichGenericIneligiblePage(taxRegime)
-        case Some(DmSpecialOfficeProcessingRequiredCESA) => whichGenericIneligiblePage(taxRegime)
+        case Some(IsAnMtdCustomer) => whichGenericIneligiblePage(taxRegime)
+        case Some(DmSpecialOfficeProcessingRequiredCESA) =>
+          if (taxRegime === Sia) throw new NotImplementedError(s"DmSpecialOfficeProcessingRequiredCESA ineligibility reason not relevant to ${taxRegime.entryName}")
```
AndyHWChung, line 71: do we know if this is the actual behaviour we want? May be worth sticking on the things to check on the other ticket if we're not sure

DreamcastKnight: it's not on the list of reasons on the ticket

DreamcastKnight: so I would have thought neither this or DmSpecialOfficeProcessingRequiredCDCS should show up for it (though as far as I can tell, DmSpecialOfficeProcessingRequiredCDCS shouldnt show up anyway but has done, for some reason?)

AndyHWChung: ah ok -  you should be able to control what reasons show up on the test-only start page. I think not all tax regimes have the same list so you could use that same mechanismt to control what's available for SIA

AndyHWChung (approved):

PR #458, opened 2024-09-23
[AC][OPS-12808-12809] pass query params to PEGA redirect & call PEGA start on WhyCannotPay

AndyHWChung:


app/controllers/WhyCannotPayInFullController.scala
```diff
@@ -81,20 +85,44 @@ class WhyCannotPayInFullController @Inject() (
...
       }
     )
   }
 
+  private def updateJourney(
+      journey:    Journey,
+      oldAnswers: Option[Set[CannotPayReason]],
+      newAnswers: Set[CannotPayReason]
+  )(implicit r: AuthenticatedJourneyRequest[_]): Future[Journey] =
+    if (oldAnswers.contains(newAnswers)) {
```
DreamcastKnight, line 106: just the smallest thing; what if the new answers have less than the old? say originally they have 2 answers and then deselect one in the change, old answers would still contain new answers, but we wouldn't register that change

AndyHWChung: the `.contains` is on the `Option` rather than on the underlying `Set`, so we should only get a `true` if the `Option` is defined and the underlying `Set`'s is equal to `newAnswers`

DreamcastKnight: ah that makes sense. cheers

DreamcastKnight (approved):

PR #462, opened 2024-10-01
[KN][OPS-12758] Update the - Terms and Conditions page

Kntanos:


app/messages/Messages.scala
```diff
@@ -1723,19 +1723,14 @@ object Messages {
       welsh   = "nad ydych yn cyflwyno’ch Ffurflenni Treth yn y dyfodol mewn pryd"
     )
 
-    val `If we cancel this agreement...`: Message = Message(
-      english = "If we cancel this agreement, you will need to pay the total amount you owe straight away.",
-      welsh   = "Os byddwn yn canslo’r cytundeb hwn, bydd yn rhaid i chi dalu’r cyfanswm sydd arnoch ar unwaith."
+    val `Contact HMRC...If we cancel this agreement...`: Message = Message(
+      english = "Contact HMRC straight away if you have a problem. If we cancel this agreement, you'll need to pay the total amount you owe. We can also use any refunds you are due to pay off your bill.",
```
AndyHWChung, line 1727: is that a straight quote in `you'll`?

app/messages/Messages.scala
```diff
@@ -1769,9 +1764,59 @@ object Messages {
...
+    )
+
+    val `Telephone...`: Message = Message(
+      english = "Telephone: <strong>0300 123 1813</strong>",
+      welsh   = "Ffôn: <strong>0300 200 1900</strong>"
+    )
+
+    val `Outside UK...`: Message = Message(
+      english = "Outside UK: <strong>+44 2890 538 192</strong>",
+      welsh   = "O’r tu allan i’r DU: +44 300 123 1813"
```
AndyHWChung, line 1779: does the welsh phone number need to be bold as well or is the difference with the english deliberate?

app/messages/Messages.scala
```diff
@@ -1769,9 +1764,59 @@ object Messages {
...
+      welsh   = "Ar gau ar benwythnosau a gwyliau banc."
+    )
+
+    val `Text service`: Message = Message(
+      english = "Text service",
+      welsh   = "Gwasanaeth Text Relay"
+    )
+
+    def `Use Relay UK...`(link: String): Message = Message(
+      english = s"""Use Relay UK if you cannot hear or speak on the telephone, dial <strong>18001</strong> then <strong>0345 300 3900</strong>. Find out more on the <a href="$link" class="govuk-link">Relay UK website (opens in new tab)</a>.""",
```
AndyHWChung, line 1803: I think we need `rel="noreferrer noopener"` for links that open in a new tab (see [here](https://github.com/hmrc/accessibility/blob/main/docs/common-issues/open-in-new-window-links.md))

app/messages/Messages.scala
Kntanos: This line is never shown in Welsh at all. I left something just in case, but I removed it now.

app/messages/Messages.scala
Kntanos: 🦅 👁️

AndyHWChung (approved):

PR #463, opened 2024-10-02
[JG] OPS-12810 - Add a language query parameter our URL's PEGA redirect to 

DreamcastKnight:


test/controllers/CanPayWithinSixMonthsControllerSpec.scala
```diff
@@ -111,15 +111,35 @@ class CanPayWithinSixMonthsControllerSpec extends ItSpec with PegaRecreateSessio
         JourneyJsonTemplates.`Obtained Can Pay Within 6 months - no`(Origins.Epaye.Bta)(testCrypto)
       )
 
-      val result = controller.canPayWithinSixMonths(Epaye)(fakeRequest)
+      val result = controller.canPayWithinSixMonths(Epaye, None)(fakeRequest)
       testPageIsDisplayed(result, Some(false))
     }
 
+    "change the language cookie to english" in {
```
AndyHWChung, line 118: it'd be worth having these sorts of tests for the other two endpoints that do the same thing as well I think so we've got full coverage

test/controllers/CanPayWithinSixMonthsControllerSpec.scala
```diff
@@ -111,15 +111,35 @@ class CanPayWithinSixMonthsControllerSpec extends ItSpec with PegaRecreateSessio
...
+      stubCommonActions()
+      EssttpBackend.CanPayWithinSixMonths.findJourney(testCrypto, Origins.Epaye.Bta)(
+        JourneyJsonTemplates.`Obtained Can Pay Within 6 months - no`(Origins.Epaye.Bta)(testCrypto)
+      )
+
+      val result = controller.canPayWithinSixMonths(Epaye, Some(English))(fakeRequest)
```
AndyHWChung, line 124: for a stronger test , we could perhaps call the controller endpoint with a fake request that had welsh in it. That way we test if language actually changes

app/controllers/CanPayWithinSixMonthsController.scala
```diff
@@ -62,8 +64,12 @@ class CanPayWithinSixMonthsController @Inject() (
               val previousAnswers = existingAnswersInJourney(request.journey)
               val form = previousAnswers.fold(CanPayWithinSixMonthsForm.form)(value =>
                 CanPayWithinSixMonthsForm.form.fill(CanPayWithinSixMonthsFormValue.canPayWithinSixMonthsToFormValue(value)))
-
-              Ok(views.canPayWithinSixMonthsPage(form, remainingAmountToPay(j)))
+              val result = Ok(views.canPayWithinSixMonthsPage(form, remainingAmountToPay(j)))
+              lang match {
+                case Some(Welsh)   => result.withCookies(Cookie("PLAY_LANG", "cy"))
+                case Some(English) => result.withCookies(Cookie("PLAY_LANG", "en"))
+                case _             => result
```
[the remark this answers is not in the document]
DreamcastKnight: oh sorry you said in other similar places, adding in other spots now

test/controllers/CanPayWithinSixMonthsControllerSpec.scala
```diff
@@ -111,15 +111,35 @@ class CanPayWithinSixMonthsControllerSpec extends ItSpec with PegaRecreateSessio
...
+      cookies(result).get("PLAY_LANG").map(_.value) shouldBe Some("en")
+    }
+
+    "change the language cookie to welsh" in {
```
AndyHWChung, line 128: we could check that the language doesn't change if `None` is passed in as well perhaps like:
```scala
Languages.values.foreach{ lang => // there's may or may not be a values you can use
  val result = controller.endpoint(regime, fakeRequestWithLang(lang))
  cookies(result).get("blah").map(_.value) shouldBe Some(lang.code)
}
```

AndyHWChung (approved):

Kntanos (approved):

PR #467, opened 2024-10-14
[KN][OPS-12847] New fields for enact API for affordability

Kntanos:


test/testsupport/testdata/TdAll.scala
```diff
@@ -471,6 +471,8 @@ object TdAll {
     ArrangementRequest(
       channelIdentifier           = ChannelIdentifiers.eSSTTP,
       regimeType                  = regimeType,
+      hasAffordabilityAssessment  = false,
```
AndyHWChung, line 474: it'd be worth having a test where this is `true` and the case ID is there as well I reckon just so it's testing that we're extracting out the right bits

AndyHWChung (approved):

PR #485, opened 2024-11-19
OPS-12937 - adding optional case id and plan recalc

DreamcastKnight:


project/SbtUpdatesSettings.scala
```diff
@@ -6,7 +6,8 @@ import sbt.Keys.*
 object SbtUpdatesSettings {
 
   lazy val sbtUpdatesSettings = Seq(
-    dependencyUpdatesFailBuild := true,
+    dependencyUpdatesFailBuild := false,
```
AndyHWChung, line 9: this may have been me - we can stick this back to `true` now I reckon

DreamcastKnight: this was actually me, my bad

AndyHWChung (approved):

PR #494, opened 2024-11-27
[TP][OPS-13007] - change error URL

tp10:


app/views/WhyCannotPayInFull.scala.html
```diff
@@ -79,7 +79,7 @@
         @govukErrorSummary(
             ErrorSummary(errorList = form.errors.map(e =>
                 ErrorLink(
-                    href = Some(s"#${e.key}"),
+                    href = Some(s"#option-UnexpectedReductionOfIncome"),
```
AndyHWChung, line 82: if the order of the items change it's quite easy to forget to update this - we could make it automatically update if we somehow base the id off of the head of `checkboxItems`

tp10: I've updated this code, thanks to @Kntanos

AndyHWChung (approved):

PR #495, opened 2024-12-02
[KN][OPS-12984] New audit event for SUPP Affordability - CanUserPayIn…

Kntanos:
…SixMonths

### DIFF
--- a/app/services/AuditService.scala
+++ b/app/services/AuditService.scala
@@ -20,16 +20,18 @@ import actionsmodel.AuthenticatedJourneyRequest
 import cats.syntax.eq._
 import essttp.bars.model.BarsVerifyStatusResponse
 import essttp.crypto.CryptoFormat
-import essttp.journey.model.Journey.{AfterCheckedPaymentPlan, AfterEnteredCanYouSetUpDirectDebit, AfterStartedPegaCase}
+import essttp.journey.model.CanPayWithinSixMonthsAnswers.CanPayWithinSixMonths
 import essttp.journey.model.Journey.Stages._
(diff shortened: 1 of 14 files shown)


app/services/AuditService.scala
```diff
@@ -126,6 +128,43 @@ class AuditService @Inject() (auditConnector: AuditConnector)(implicit ec: Execu
...
+  def auditCanUserPayInSixMonths(journey: Journey, canPay: CanPayWithinSixMonths)(implicit hc: HeaderCarrier): Unit = {
+    audit(toCanUserPayInSixMonths(journey, canPay))
+  }
+
+  private def toCanUserPayInSixMonths(journey: Journey, canPay: CanPayWithinSixMonths): CanUserPayInSixMonthsAuditDetail = {
+    CanUserPayInSixMonthsAuditDetail(
+      regime             = journey.taxRegime.entryName,
+      taxIdentifier      = taxIdentifierToAudit(journey),
+      pegaCaseId         = journey.pegaCaseId,
+      correlationId      = journey.correlationId,
```
AndyHWChung, line 140: do we need another field for the correlation id we sent to PEGA in the start case API request? In another audit event for the get case response we added in `pegaCorrelationId` which in the ticket was originally just `correlationId` but we changed it cos it was confusing with the existing `correlationId` you've made use of here

Kntanos: 🤷🏻  Do you recon that this correlationId refers to the pegaCorrelationId, or that they missed including in?

AndyHWChung (approved):

PR #496, opened 2024-12-04
[KN][OPS-13083] Check your payment plan page and Why are you unable t…

Kntanos:
…o pay in full? amendment

### DIFF
--- a/app/views/checkpaymentchedule/CheckPaymentScheduleRows.scala
+++ b/app/views/checkpaymentchedule/CheckPaymentScheduleRows.scala
@@ -16,7 +16,6 @@
 
 package views.checkpaymentchedule
 
-import cats.implicits.catsSyntaxEq
 import essttp.journey.model.{CanPayWithinSixMonthsAnswers, UpfrontPaymentAnswers, WhyCannotPayInFullAnswers}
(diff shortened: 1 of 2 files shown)


tp10 (approved):

app/views/checkpaymentchedule/CheckPaymentScheduleRows.scala
```diff
@@ -31,9 +32,11 @@ object CheckPaymentScheduleRows {
       changeLinkCall:            Call
   )(implicit lang: Language): Option[SummaryListRow] = {
       def bullets(reasons: Set[CannotPayReason]) = {
+        val htmlClass = if (reasons.size === 1) "govuk-list" else "govuk-list govuk-list--bullet"
+
         Html(
           s"""
-           |<ul class="govuk-list govuk-list--bullet">
+           |<ul class="$htmlClass">
```
AndyHWChung, line 39: if `reason.size` is one we probably don't need the `<ul>` right?

AndyHWChung: it might be easier to `toList` the set and pattern match to make it easier to deal with the size one case

AndyHWChung (approved):

PR #504, opened 2024-12-17
[JG] OPS-12893 - Simple Assessment update details kick out

DreamcastKnight:


app/views/partials/GenericRLSPartial.scala.html
```diff
@@ -35,4 +35,4 @@
 
 <p id="update-details-content" class="govuk-body">@{Html(Messages.NotEligible.`You must update your details with HMRC`(tellHMRCChangeDetailsLink).show)}</p>
 
-<p id="call-us-content" class="govuk-body">@{Html(Messages.NotEligible.`Call us on 0300 123 1813 as you may be able to set up a plan over the phone`.show)}</p>
+<p id="call-us-content" class="govuk-body">@{Html(Messages.NotEligible.`Call us on 0300 322 7835 as you may be able to set up a plan over the phone`.show)}</p>
```
AndyHWChung, line 38: just to double check, is the phone number changing for all the regimes? If it is, is the old message still being used?

conf/app.routes
```diff
@@ -36,7 +36,7 @@ GET    /determine-eligibility                               controllers.Determin
 GET    /not-enrolled                                        controllers.NotEnrolledController.notEnrolled
 GET    /not-vat-registered                                  controllers.NotEnrolledController.notVatRegistered
 GET    /request-access-to-self-assessment                   controllers.NotEnrolledController.notSaEnrolled
-GET    /no-nino-found-sia                                   controllers.NotEnrolledController.simpNoNino
+GET    /no-nino-found-simp                                  controllers.NotEnrolledController.simpNoNino
```
AndyHWChung, line 39: this is going to disappear eventually anyway I think so we don't technically need this change

app/views/partials/GenericRLSPartial.scala.html
DreamcastKnight: if it's changing for one, I don't see why it wouldn't for the others; though I have now realised that the welsh needs to be 0300 200 1900 according to the spec so will change that, at least

conf/app.routes
DreamcastKnight: fair

test/controllers/IneligibleControllerSpec.scala
```diff
@@ -330,7 +330,7 @@ class IneligibleControllerSpec extends ItSpec {
                   "You must <a href=\"https://www.gov.uk/tell-hmrc-change-of-details\" class=\"govuk-link\">update your details with HMRC</a>. After you’ve updated your details, wait 3 working days before trying again online.")
               }
 
-              val expectedCallUsContent = "Call us on <strong>0300 322 7835</strong> as you may be able to set up a plan over the phone."
+              val expectedCallUsContent = "Call us on <strong>0300 123 1813</strong> as you may be able to set up a plan over the phone."
```
AndyHWChung, line 333: is the SIMP RLS ineligible case being tested in this spec? It may need adding in - I'd expect you to have to change the expected text here based on the tax. regime

AndyHWChung (approved):

PR #520, opened 2025-01-24
[KN][OPS-12233] Deal with HMRC-MTD-IT enrolment - CR134

Kntanos:


test/controllers/DetermineEligibilityControllerSpec.scala
```diff
@@ -739,5 +747,34 @@ class DetermineEligibilityControllerSpec extends ItSpec with CombinationsHelper
       redirectLocation(result) shouldBe Some(PageUrls.epayeDebtTooSmallUrl)
     }
 
+    "SA user with IR-SA enrolment but no NINO found, should be directed to generic ineligible page" in {
+      val eligibilityRules = TdAll.eligibleEligibilityRules.copy(part2 = eligibleEligibilityRules.part2.copy(noMtditsaEnrollment = Some(false)))
```
AndyHWChung, line 751: in reality we wouldn't get `noMtditsaEnrollment` returned to us in `Ttp.Eligibility.stubRetrieveEligibility`

app/testOnly/LoginRequestMaker.scala
```diff
@@ -92,4 +93,15 @@ object LoginRequestMaker {
     "state" -> irSaEnrolment.enrolmentStatus.toString
   )
 
+  private def makeIrSaAndMtdItEnrolmentJson(mtdItEnrolment: MtdItEnrolment): JsObject = Json.obj(
+    "key" -> "HMRC-MTD-IT",
+    "identifiers" -> Json.arr(
+      Json.obj(
+        "key" -> "MTDITID",
+        "value" -> mtdItEnrolment.nino.value
```
AndyHWChung, line 101: might not make too much sense to have `nino` in the `MtdItEnrolment` case class anymore - it'd make sense to have a separate `MtdItId` field on the test-only form which when left blank will just result in a random id being generated. Not sure what the form of the id is but if we generate 8 random digits or something then that should be good enough

AndyHWChung (approved):

PR #535, opened 2025-02-19
[JG] OPS-13314 - amending sa payment plan set up page

DreamcastKnight:


app/views/SaPaymentPlanSetUpPage.scala.html
```diff
@@ -107,7 +107,7 @@ <h2 class="govuk-heading-m">
 
 
     <p class="govuk-body">@{
-        Messages.Confirmation.`Your tax account will be updated with your payment plan within 24 hours.`.show
+        Messages.Confirmation.`HMRC records will be updated with your payment plan details within 24 hours.`.show
```
AndyHWChung, line 110: doesn't the stuff below this need to change as well? If you look at the other plan set up page then you'll see the content already there that you can re-use

AndyHWChung (approved):

PR #543, opened 2025-03-21
[KN][OPS-13361] Correct sign out for OneLogin

Kntanos:


app/controllers/SignOutController.scala
```diff
@@ -39,18 +37,20 @@ class SignOutController @Inject() (
   def signOutFromTimeout: Action[AnyContent] = Action { implicit request =>
     // N.B. the implicit request being passed into the page here may still have the auth
     // token in it so take care to ensure that the sign out link is not shown by mistake
-    Ok(views.timedOutPage()).withNewSession
+    val continueUrl = routes.SignOutController.signOutFromTimeout.absoluteURL()
```
AndyHWChung, line 40: won't this result in an infinite redirect loop? You may need another controller endpoint which just serves the `views.timedOutPage` content to redirect to here

app/controllers/SignOutController.scala
```diff
@@ -39,18 +37,20 @@ class SignOutController @Inject() (
...
-  def signOut: Action[AnyContent] = as.authenticatedJourneyAction { implicit request =>
-    Redirect {
-      request.journey.taxRegime match {
-        case TaxRegime.Epaye => routes.SignOutController.exitSurveyPaye
-        case TaxRegime.Vat   => routes.SignOutController.exitSurveyVat
-        case TaxRegime.Sa    => routes.SignOutController.exitSurveySa
-        case TaxRegime.Simp  => routes.SignOutController.exitSurveySimp
-      }
-    }.withNewSession
+  def signOut(): Action[AnyContent] = as.authenticatedJourneyAction { implicit request =>
```
AndyHWChung, line 45: this and the other controller endpoints here could just be a `val`'s

app/controllers/SignOutController.scala
Kntanos: oh yeah, but it will be spectacular
..will fix

PR #549, opened 2025-04-02
[KN][OPS-13483] Add Text to your Bill page for Simple Assessment

Kntanos:


test/controllers/YourBillControllerSpec.scala
```diff
@@ -241,6 +241,42 @@ class YourBillControllerSpec extends ItSpec {
...
+
+      tableRows(0)
+        .select(".govuk-summary-list__key")
+        .text() shouldBe "13 Jul 2020 to 14 Jul 2020 Bill due 7 February 2017"
+      tableRows(0).select(".govuk-summary-list__value").text() shouldBe "£2,000 (includes interest added to date)"
+
+      tableRows(1).select(".govuk-summary-list__key").text() shouldBe "13 Aug 2020 to 14 Aug 2020 Bill due 7 March 2017"
+      tableRows(1).select(".govuk-summary-list__value").text() shouldBe "£1,000 (includes interest added to date)"
+
+      val extraText = doc.select(".govuk-body").asScala.toSeq
```
AndyHWChung, line 272: is there anything you could stick into the tests for the other regimes that will fail if the new text accidentally shows up for them?

test/controllers/YourBillControllerSpec.scala
```diff
@@ -241,6 +257,42 @@ class YourBillControllerSpec extends ItSpec {
...
```
AndyHWChung, line 288: you could use your new id's here now to make the tests look a bit more more consistent. If you're gonna be making changes then maybe changing `#extra-para1` to something like `#simp-extra-para1` would make it clearer what it is and future-proof it a bit

AndyHWChung (approved):

PR #558, opened 2025-05-14
[JG] OPS-13646 - updating interest rate

DreamcastKnight:


Kntanos (approved):

test/controllers/LandingPageControllerSpec.scala
```diff
@@ -107,7 +107,7 @@ class LandingPageControllerSpec extends ItSpec {
 
       paragraphs(0)
         .text() shouldBe "Use this service to set up a payment plan for your outstanding employers’ PAYE bill. " +
-        "Payments are taken by Direct Debit and include interest charged at the Bank of England base rate plus 2.5% per year."
+        "Payments are taken by Direct Debit and include interest charged at the Bank of England base rate plus 4.0% per year."
```
AndyHWChung, line 110: shouldn't this just say `4%` instead of `4.0%`? You may need to do some formatting if it's a whole number to make that happen

DreamcastKnight: it should do, I think, but im unsure what I need to do to make it do that

app/messages/Messages.scala
```diff
@@ -1642,9 +1642,9 @@ object Messages {
       welsh = "Rydym yn codi llog ar bob swm sy’n hwyr."
     )
 
-    val `We charge the Bank of England base rate plus 2.5% per year`: Message = Message(
-      english = "We charge the <strong>Bank of England base rate plus 2.5%</strong> per year.",
-      welsh = "Rydym yn codi <strong>cyfradd sylfaenol Banc Lloegr ynghyd â 2.5%</strong> y flwyddyn."
+    val `We charge the Bank of England base rate plus 4% per year`: Message = Message(
```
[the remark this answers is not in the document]
DreamcastKnight: can do

test/controllers/LandingPageControllerSpec.scala
AndyHWChung: it could be something as simple as this maybe:
```scala
def formatPercentage(d: BigDecimal): String = d.toString.stripSuffix(".0")
```

AndyHWChung (approved):

PR #613, opened 2026-03-24
[AC]OPS-14742] allow optional SAFEID for SA testonly

AndyHWChung:


DreamcastKnight (commented):
is it worth throwing in a quick test somewhere for when there is/isnt a safe id?

AndyHWChung:
> is it worth throwing in a quick test somewhere for when there is/isnt a safe id?

I could add in the safe id to some tests but it'd be a bit artificial cos we don't test from eligibility at the start right through to enact in the same test. I'd either add it to the eligibility tests or the enact tests (cos we don't really have tests for the test-only stuff) but we're already kind of proving there that we're transfer whatever we get in the eligibility identification array to the session object and likewise we're kind of already testing we transfer the eligibility identification request to the enact request body. I've tested this safe ID stuff manually at least.

DreamcastKnight (approved):

PR #615, opened 2026-04-15
OPS-14954 - adding check for nino in identification

DreamcastKnight:


app/services/TtpService.scala
```diff
@@ -157,9 +157,11 @@ class TtpService @Inject() (
 
     val identification = taxRegime match {
       case TaxRegime.Sa =>
-        val maybeNinoId =
-          authenticatedJourneyRequest.nino.map(nino => Identification(IdType("NINO"), IdValue(nino.value)))
-        maybeNinoId.fold(eligibilityCheckResult.identification)(_ :: eligibilityCheckResult.identification)
+        if (eligibilityCheckResult.identification.map(_.idType).contains("NINO")) eligibilityCheckResult.identification
```
AndyHWChung: it'd be nice to have a unit test for the scenario when there's already a NINO to make a duplicate one doesn't appear

barnesjake (approved):

PR #627, opened 2026-05-18
15037

JamieNorth:


app/messages/Messages.scala
```diff
@@ -939,8 +939,8 @@ object Messages {
     )
 
     val `Select all that apply or 'none of these'` : Message = Message(
-      english = "Select all that apply or ‘None of these’",
-      welsh = "Dewiswch bob un sy’n berthnasol neu ‘Dim un o’r rhain’"
+      english = "Select why you are unable to pay in full, or select 'None of these'",
```
AndyHWChung: these should be curly quotes around `None of these`

JamieNorth: 👍

AndyHWChung (approved):

PR #628, opened 2026-05-19
[JN][OPS-15058] - Adding h2 header on Check Your Payment Plan page

JamieNorth:


app/views/checkpaymentchedule/CheckPaymentSchedule.scala.html
```diff
@@ -119,6 +119,10 @@ <h1 class="govuk-heading-l">
     @title
   </h1>
 
+<h2 class="govuk-heading-m">
```
AndyHWChung, line 122: I suspect we're testing the content on this page - it'd be good to add something in there to test this heading is appearing

JamieNorth: Did check this, looks like only testing content of tables rather than headings.

can have a look at adding tests for the headings too, although I suspect headings in general haven’t been asserted against on other pages

AndyHWChung: oh yeah - looks like we;re not testing the h2's. One approach might be to use jsoup to get the previous element just before each summary list to check it's a h2 with the right content and class

JamieNorth: Added, thanks :)

AndyHWChung (approved):

PR #629, opened 2026-05-22
[JG] OPS-15059 -  Overly long hint text 

DreamcastKnight:
- changing why cannot pay in full page

### DIFF
--- a/app/messages/Messages.scala
+++ b/app/messages/Messages.scala
@@ -928,9 +928,43 @@ object Messages {
 
   object WhyCannotPayInFull {
 
-    val `Why are you unable to pay in full?` : Message = Message(
-      english = "Why are you unable to pay in full?",
-      welsh = "Pam nad oes modd i chi dalu’ch llawn?"
+    val `Tell us why you cannot pay in full`: Message = Message(
+      english = "Tell us why you cannot pay in full",
+      welsh = "Rhowch wybod i ni pam na allwch dalu’r swm llawn"
+    )
+
+    val `Which of these apply to you?` : Message = Message(
+      english = "Which of these apply to you?",
+      welsh = "Pa un o’r o’r rhain sy’n berthnasol i chi?"
+    )
+
+    val `Reasons could include`: Message = Message(
+      english = "Reasons could include:",
+      welsh = "Gall y rhesymau gynnwys:"
+    )
+
+    val `lost or reduced income, business or employment-1`: Message = Message(
+      english = "lost or reduced income, business or employment",
+      welsh = "gostyngiad mewn incwm, busnes neu gyflogaeth"
+    )
+    val `lost or reduced income, business or employment-2`: Message = Message(
+      english = "",
+      welsh = "colli incwm, busnes neu gyflogaeth"
+    )
+
+    val `unexpected costs such as repairs following theft or damage`: Message = Message(
+      english = "unexpected costs such as repairs following theft or damage",
+      welsh = "costau annisgwyl fel atgyweiriadau yn dilyn lladrad neu ddifrod"
+    )
+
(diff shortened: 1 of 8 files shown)


app/messages/Messages.scala
AndyHWChung: if there's no hint text now with the checkboxes we can just return `Message` here without an `Option[Message]`

test/controllers/WhyCannotPayInFullControllerSpec.scala
```diff
@@ -69,13 +69,13 @@ class WhyCannotPayInFullControllerSpec extends ItSpec, UnchangedFromCYALinkAsser
         CheckBoxInfo(
           "UnexpectedReductionOfIncome",
           "Unexpected reduction of income",
-          "For example, lost or reduced business or unemployment.",
+          "",
```
AndyHWChung: if hint text part of `CheckBoxInfo` can go now if there's no hint text anymore

app/views/WhyCannotPayInFull.scala.html
AndyHWChung: this needs to change as well to `"Select all that apply"`

test/controllers/WhyCannotPayInFullControllerSpec.scala
AndyHWChung, line 60: we should have unit tests for the new sentence with the bullet points as well - catering somehow for the extra bullet point in welsh. There's a bullet list on the start pages if you need inspiration on the tests

test/testsupport/ItSpec.scala
```diff
@@ -111,7 +111,9 @@ class ItSpec extends UnitSpec, GuiceOneServerPerSuite, WireMockSupport, HttpRead
     }
   }
 
-  val fakeRequest = FakeRequest().withAuthToken().withSession(SessionKeys.sessionId -> "IamATestSessionId")
+  val fakeRequest      = FakeRequest().withAuthToken().withSession(SessionKeys.sessionId -> "IamATestSessionId")
+  val fakeRequestWelsh =
```
AndyHWChung: we can already do `fakeRequest.withLangWelsh()` to avoid having this extra `val`

test/controllers/WhyCannotPayInFullControllerSpec.scala
```diff
@@ -150,6 +153,27 @@ class WhyCannotPayInFullControllerSpec extends ItSpec, UnchangedFromCYALinkAsser
       testPageIsDisplayed(result, TdAll.whyCannotPayReasons)
     }
 
+    "display the page with correct bullet points in welsh" in {
```
AndyHWChung, line 177: there's already a `"display the page in Welsh"` test in this spec you can add to to make use of the test set up that's already there

DreamcastKnight: is there? I can't find any reference to welsh in the file at all

AndyHWChung: I think I was looking at the wrong spec 😅 in that case then it'd probably be worth adding one in - the easiest way would be to add a language parameter to `testPageIsDisplayed` in the existing spec and adjust accordingly. I think you can call `fold` on the language to make it a bit easier

DreamcastKnight: or maybe making a "testpageIsDisplayedWelsh" would make more sense?

AndyHWChung: I was thinking if you reuse it then it'd avoid having a lot of the stuff copy and pasted - at least the framework is already there to extract out all the necessary bits

AndyHWChung (approved):

PR #635, opened 2026-07-07
[JG] OPS-14997 - changes for new schema

DreamcastKnight:


test/controllers/DetermineEligibilityControllerSpec.scala
```diff
@@ -538,60 +599,27 @@ class DetermineEligibilityControllerSpec extends ItSpec, CombinationsHelper {
...
           "ChargesBeforeMaxAccountingDate - VAT",
+          TdAll.eligibleEligibilityRules,
           TdAll.notEligibleChargesBeforeMaxAccountingDate,
           "chargesBeforeMaxAccountingDate",
           PageUrls.vatDebtBeforeAccountingDateUrl,
           JourneyJsonTemplates.`Eligibility Checked - Ineligible - ChargesBeforeMaxAccountingDate`(Origins.Vat.Bta),
           Origins.Vat.Bta
-        ),
-        (
-          "HasDisguisedRemuneration - SA",
```
AndyHWChung, line 555: these tests should still be here shouldn't they?

app/models/EligibilityError.scala
```diff
@@ -79,35 +79,44 @@ object EligibilityErrors extends Enum[EligibilityError] {
...
-      case e if e.noMtditsaEnrollment.contains(true)                   => Some(NoMtditsaEnrollment)
-      case e if e.allChargeTypeAssessmentsFailed.contains(true)        => Some(AllChargeTypeAssessmentsFailed)
-      case e if e.noValidPlanAfterAssessments.contains(true)           => Some(NoValidPlanAfterAssessments)
-      case _                                                           => None // all false
+  def toEligibilityError(
+    eligibilityRules:           EligibilityRules,
+    assessmentEligibilityRules: AssessmentEligibilityRules
+  ): Option[EligibilityError] =
+    (eligibilityRules, assessmentEligibilityRules) match {
+      case (e1, e2) if e1.atLeastOneReasonForIneligibility && e2.atLeastOneReasonForIneligibility =>
```
AndyHWChung, line 87: it'd be good to get a test for the three cases which map to `MultipleReasons` - one where each eligibility rules has one and one each for when each has more than one but the other has none. Not sure if there's such a test in `DetermineEligibilityControllerSpec` yet

AndyHWChung (approved):

PR #636, opened 2026-07-15
[JG] OPS-15045 - adding futureLiabilitiesExcluded feature flag

DreamcastKnight:


conf/application.conf
```diff
@@ -165,6 +165,7 @@ features {
   sa = true
   simp = true
   eligibilityReqIdentificationFlag = true
+  futureLiabilitiesExcluded = true
```
AndyHWChung: could we make this a bit more general like `future-dated-liabilities`?

app/views/simp/SimpLanding.scala.html
```diff
@@ -47,6 +47,9 @@
   <h1 class="govuk-heading-l">@{Messages.Simp.`Set up a Simple Assessment payment plan`.show}</h1>
 
   <p class="govuk-body">@{Messages.Simp.`You can use this service to pay overdue payments in instalments.`.show}</p>
+@if(appConfig.futureLiabilitiesExcluded) {
```
AndyHWChung: we should have a test for this to check if the content shows if enabled and doesn't show if not enabled

conf/application.conf
AndyHWChung: also it'd be good to raise a PR in app-config-prod now to make sure it's definitely set to false there so we don't forget

app/messages/Messages.scala
```diff
@@ -1392,6 +1392,12 @@ object Messages {
       welsh = "Gallwch ddefnyddio’r gwasanaeth hwn i dalu taliadau hwyr fesul rhandaliad."
     )
 
+    val `Additionally, if you have a Simple Assessment tax bill that is not yet due, you can pay this in advance instalments.`
+      : Message = Message(
+      english =
```
AndyHWChung, line 1397: do we not have the welsh?

DreamcastKnight: I dont think so; nothing in the ticket, I was wondering about that

app/config/AppConfig.scala
```diff
@@ -35,7 +35,7 @@ class AppConfig @Inject() (config: Configuration, servicesConfig: ServicesConfig
   val emailJourneyEnabled: Boolean                                         = config.get[Boolean]("features.email-journey")
   val saEnabled: Boolean                                                   = config.get[Boolean]("features.sa")
   val simpEnabled: Boolean                                                 = config.get[Boolean]("features.simp")
-  val futureLiabilitiesExcluded: Boolean                                   = config.get[Boolean]("features.futureLiabilitiesExcluded")
+  val future_dated_liabilities: Boolean                                    = config.get[Boolean]("features.future-dated-liabilities")
```
AndyHWChung: it'd be nice to follow the style of the other flags we've got like `futureDatedLiabiltiesEnabled`

AndyHWChung (approved):

PR #645, opened 2026-08-26
[JN][OPS-15355] - Additional hint text

JamieNorth:


app/messages/Messages.scala
```diff
@@ -287,6 +287,12 @@ object Messages {
       welsh = "Ydych chi am ychwanegu’r bil treth hwn sydd i ddod at eich cynllun talu?"
     )
 
+    val `If you do not add this payment to your plan`: Message = Message(
+      english = "If you do not add this payment to your plan, you must pay it by the due date shown.",
+      welsh =
+        "Os na fyddwch yn ychwanegu'r taliad hwn at eich cynllun, mae’n rhaid i chi ei dalu erbyn y dyddiad dyledus a ddangosir."
```
AndyHWChung: should be a curly quote instead of a straight quote in `ychwanegu'r`

app/views/AdvancePayments.scala.html
```diff
@@ -81,6 +81,7 @@ <h1 class="govuk-heading-l">@title</h1>
             )),
             idPrefix = Some(key),
             name = key,
+            hint = Some(Hint(content = Text(Messages.AdvancePayments.`If you do not add this payment to your plan`.show))),
```
AndyHWChung, line 84: there should be a test for this page that already exists that we can add to to test for the hint text

app/messages/Messages.scala
JamieNorth: Curly added

app/views/AdvancePayments.scala.html
JamieNorth: Assertion added for hint text as part of existing test

AndyHWChung (approved):

